PR #29, opened 2014-03-11
Enhancement and IntelliJ project files

GlenCoakley:
Dimitry etc.,

I have a pull request for you do you with two logical sets of changes.  Do you prefer more requests with smaller changes or fewer requests with more changes in them?

I made a minor change to extend the filtering to also look at attributes.  I wasn't sure of the procedure for translations.  It looks like there is some kind of tool.  But, there are also may resource keys that are not in every language-specific resource*.properties file.  I used Google translate to translate the one I added. 

In a separate commit I also added a .idea directory off the root and a *.iml file in each project root directory.  These files provide IntelliJ IDEA project configuration.  I switched from Eclipse to IDEA over a year ago and like it much better particularly for the large source code base I have where I work.   IDEA came out with a free "Community Edition" a few years ago which can be downloaded from here: 
http://www.jetbrains.com/idea/features/editions_comparison_matrix.html

I was on the forums hunting for information for developers and agree with some of the comments there.  I think the project will gather more help if there is a Contributer section of a wiki that would provide high-level information about the project for people that would like to help.  It can include the state of tests, how translations should are done, that there are dependency issues and the latest 1.3 branch seem to have compile errors, etc.  A forum is good for discussion but, very tedious to try and find information in one.


dpolivaev:
This PR is closed as just as too old.
Feel welcome to submit new pull requests if you like.

PR #44, opened 2015-03-08
better mm2html.xsl

hokorobi:
fix indent
output image

### DIFF
--- a/freeplane/external-resources/xslt/mm2html.xsl
+++ b/freeplane/external-resources/xslt/mm2html.xsl
@@ -1,139 +1,149 @@
 <?xml version="1.0" encoding="UTF-8" ?>
 
-	<!--
-		MINDMAPEXPORTFILTER html;htm %xslt_export.html
-		
-		: This code released under the GPL.
-		: (http://www.gnu.org/copyleft/gpl.html) Document : mindmap2html.xsl
-		Created on : 01 February 2004, 17:17 Author : joerg feuerhake
-		joerg.feuerhake@free-penguin.org Description: transforms freeplane mm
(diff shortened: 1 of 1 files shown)


fnatter:
hello hokorobi,

>fix indent
>output image

thanks for your contribution to Freeplane.
Your changes look ok to me, but there is one thing:

    < 		MINDMAPEXPORTFILTER html;htm %xslt_export.html
    ---
    >     MINDMAPEXPORTFILTER html;htm HTML with image

'%xslt_export' is a translation key (results in i.e. "Html Dokument" in
German or "Documento HTML" in Spanish or ...). Your change has hardcoded
the "HTML with image" export filter name.

There are two solutions:

(1) Revert this change (i.e. keep '%xslt_export.html')

(2) If you want to create a separate exporter (like when there are use cases
when the user doesn't want to export images which I cannot think of):
- put the changes in mm2html_with_images.xsl and
- use MINDMAPEXPORTFILTER html;htm %xslt_export.html_with_images
- and add translations for xslt_export.html_with_images

What do you think?

Thanks and Best Regards,
Felix

hokorobi:
Hello, Felixb.

Thank you for your comments.
I employ "(1)".

Best regards,
hokorobi

fnatter:
hello hokorobi,

will you create a new pull request?

Best Regards,
Felix

dpolivaev:
Usually new pull requests are not needed because all commits pushed to the branch by the request author are automatically added to the original request.

dpolivaev:
So for instance this pull request consists of two commits, the original one and the new one

hokorobi:
Thank you

PR #58, opened 2016-11-20
Add initial editorconfig

koppor:
I am working with IntelliJ, which uses editorconfig to keep the files consistent.

I saw that in `*.java`, tab is used. In the other files (e..g, `build.gradle`), there is a mix between tabs and spaces. I don't know, where the project aims for in this regard, but this editorconfig can be a good start to harmonize tabs and spaces in the project.

For more infomration on editorconfig see http://editorconfig.org/

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,8 @@
+root = true
+
+[*]
+charset = utf-8
+insert_final_newline = true
+
+[*.java]
+indent_style = tab


dpolivaev:
We usually use eclipse because of our use of OSGi framework for managing freeplane parts.
It has its own formatter configuration which is not applied to the top level directory, because eclipse projects are placed a level below. I think we can add your configuration and try to add plug-ins to our tools. Let me discuss with the other devs.

vboerchers:
Even if we have the formatter since the beginning ([freeplane_framework/eclipse/CodeStyle](https://github.com/freeplane/freeplane/tree/1.5.x/freeplane_framework/eclipse/CodeStyle)) and every dev is using eclipse even new code sometime doesn't follow the rules. In eclipse we should use [save actions](http://help.eclipse.org/mars/index.jsp?topic=%2Forg.eclipse.jdt.doc.user%2Freference%2Fpreferences%2Fjava%2Feditor%2Fref-preferences-save-actions.htm) to apply the formatter either to the whole file or to new code. I guess Idea has something similar.

koppor:
IntelliJ supports save actions only using a plugin:
http://stackoverflow.com/a/32137623/873282. By using the editorconfig, it
is ensured that also during editing, some parts of the formatting are
correct. Having an additional editorconfig does not harm IMHO.

IntelliJ seems to support OSGi, too:
https://www.jetbrains.com/help/idea/2016.2/osgi.html - I used IntelliJ to
quickly edit the freeplane code without a long setup.

Side comment: To force each developer to have the same eclipse config, we
let gradle generate the Eclipse config:
https://github.com/JabRef/jabref/blob/master/eclipse.gradle and do not rely
on the users to import the settings.

koppor:
There is an eclipse plugin for EditorConfig support. https://marketplace.eclipse.org/content/editorconfig-eclipse. I still think that `.editorconfig` is **the** way to specify requirements on the files.

PR #60, opened 2016-11-20
Update JGoodies

koppor:
Debian updated jgoodies forms from 1.6.0 to 1.9.0 (https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=819669) and jgoodies common from 1.4.0 to 1.8.1 (https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=819668). The changelog of jgoodies-forms 1.9.0 reads:

    This version is binary incompatible with previous versions,
    primarily because deprecated code has been removed.
    Migration is necessary (but easy) only for those API users
    that have used the open source ListViewBuilder instead of
    the commercial ListViewBuilder, and that have used deprecated
    methods in PanelBuilder and DefaultFormBuilder.

Thus, freeplane 1.5.1 does not build anymore. Fortunately, to make it compiling again, only a small patch is required. This patch is provided here.

### DIFF
--- a/freeplane/build.gradle
+++ b/freeplane/build.gradle
@@ -12,8 +12,8 @@ dependencies {
     compile 'commons-lang:commons-lang:2.6',
             'commons-io:commons-io:2.4',
             'commons-codec:commons-codec:1.7',
-            'com.jgoodies:jgoodies-forms:1.6.0',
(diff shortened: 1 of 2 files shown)


dpolivaev:
Felix, could you please check this suggestion in relation to debian?

Issue #48, opened 2016-01-15
how to setup Freeplane for development with Intellij / gradle?


gotohr:
I've checked out 1.5.x branch.
What should I do next?

dpolivaev:
To build Freeplane with gradle just execute `gradle build`.

For developing with eclipse you can execute `gradle clean build cleanEclipse eclipse`.
which generates eclipse project files and osdgi manifest files which can be used specially running freeplane as OSGi application under eclipse.

We do not know how to develop freeplane with IntelliJ. Please tell us if you find out how to do it.

dpolivaev:
For eclipse there are also launch configuration files provided.

aadrian:
If the project is a "clean" multi-module Gradle project (without dynamic hacks), than IntelliJ can import it and keep the settings in sync. Just use "Import Project" and in the wizard allow to scan recursively.

dpolivaev:
If you find out how to develop and debug Freeplane using IntelliJ please share it with us. We tried and failed.

aadrian:
> We tried and failed.

What exactly was not working for you? 

To try it, I just imported a fresh project and IntelliJ (Ultimate) seems to recognize so far 13 Gradle modules as IntelliJ modules.
![freeplane_intellij_1](https://cloud.githubusercontent.com/assets/140953/12364137/427dbf60-bbcd-11e5-9da4-2f7a77e2556a.png)

Of course, the IntelliJ specific files are not in ```.gitignore``` yet:
![freeplane_git_1](https://cloud.githubusercontent.com/assets/140953/12364223/b7d46a98-bbcd-11e5-9694-afb1e17e1902.png)

dpolivaev:
Running a debug session with Freeplane as OSGi - Application 
(Freeplane is deployed with knopflerfish, http://www.knopflerfish.org/)

dpolivaev:
It should not matter which OSGi framework you take

dpolivaev:
Developing without debug possibility is not really comfortable, therefore we had to use only eclipse until now.

aadrian:
> Running a debug session with Freeplane as OSGi 

Just curious, what's the advantage of OSGi for a tool like FreePlane?

Shouldn't the entire application just work by just starting ```FreeplaneMain``` ?
![freeplane_intellij_2](https://cloud.githubusercontent.com/assets/140953/12365170/6c65dae6-bbd3-11e5-8059-ea6c4088b874.png)

[...]

PR #62, opened 2016-12-03
Add gradle wrapper to project

inglor:
Gradle wrapper removes one step from developer's setup and Continuous Integration, more info on [docs]

With this the gradle dependency for distribution packages (-git versions) would be removed [freeplane-git Archlinux package]

[docs]: https://docs.gradle.org/current/userguide/gradle_wrapper.html
[freeplane-git Archlinux package]: https://aur.archlinux.org/packages/freeplane-git/


dpolivaev:
Why should we add gradle wrapper? Nobody commits ant, maven or jdk binaries, so why is gradle different?

inglor:
This would avoid questions such as #48 and as explained on the description the VCS packages for archlinux (not sure about debian) currently needs gradle as a make dependency. Apart from these, quoting the gradle docs:

> Most tools require installation on your computer before you can use them. If the installation is easy, you may think that’s fine. But it can be an unnecessary burden on the users of the build. Equally importantly, will the user install the right version of the tool for the build? What if they’re building an old version of the software?
The Gradle Wrapper (henceforth referred to as the “Wrapper”) solves both these problems and is the preferred way of starting a Gradle build.

fnatter:
I am also not sure that this is helpful (unless gradlew would automatically use the gradle build
daemon, which would improve build times).
Debian/Ubuntu is not an argument, as gradle needs to be packaged separately, and cannot be downloaded.

Cheers and Best Regards,
Felix

inglor:
@fnatter The gradle 3+ automatically enabled gradle daemon. The recommendation though for CI build is to have it disabled [1]

[1]: https://docs.gradle.org/current/userguide/gradle_daemon.html#when_should_i_not_use_the_gradle_daemon

inglor:
PS: maven has a wrapper https://github.com/takari/maven-wrapper

inglor:
Bump, this has been rebased against 1.6.x and using latest gradle which is much faster in incremental builds

PR #64, opened 2016-12-04
First set of control groups plus some changes per suggestions by Dimitry.

joetopshot:
First set of control groups plus some changes per suggestions by Dimitry.


joetopshot:
Your suggestion about creating a ControlGroup array in StyleEditorPane will not work because the items to be displayed in the panel need to be in a particular order, including horizontal lines, etc.  I therefore created a map of ControlGroup items which will give us both the individual ControlGroup as needed in getControls() as well as the ability to cycle through the whole list.

dpolivaev:
Separators can be treated like ControlGroup as follows.
The order of controls and separators can given by the order of elements in the array.
The array can be injected e.g. as constructor parameter.
This would solution make StyleEditorPanel independent of the controls and separators it contains.

```java
class GroupSeparator implements ControlGroup{
	final String labelKey;

	public GroupSeparator(String labelKey) {
		this.labelKey = labelKey;
	}

	@Override
	public void setStyle(NodeModel node) {
		// intentionally left blank
	}

	@Override
	public void addControlGroup(List<IPropertyControl> controls) {
		controls.add(new SeparatorProperty(labelKey));
	}
}
```

dpolivaev:
You could also consider use of composite pattern where  ControlGroup contains a separator and all related control groups as its internal elements.

```java
class CompositeControlGroup implements ControlGroup{
	
	private final ControlGroup[] elements;

	public CompositeControlGroup(ControlGroup[] elements) {
		super();
		this.elements = elements;
	}

	@Override
	public void setStyle(NodeModel node) {
		for(ControlGroup element:elements)
			element.setStyle(node);
	}

	@Override
	public void addControlGroup(List<IPropertyControl> controls) {
		for(ControlGroup element:elements)
			element.addControlGroup(controls);
	}
}
```

dpolivaev:
Looking for a better name for all our listeners (ControlGroupChangeListener and its derivated classes) I have got an idea of ControlGroupChangePropagator because actually it propagates changes from control to the map and from the map to the control. All member variables could be renamed accordingly.

What do you think about this idea?

joetopshot:
Regarding the name ControlGroupChangeListener versus your ControlGroupChangePropagator, I personally like ControlGroupChangeListener better.  It is an event listener and having "Listener" in the name tells us that.  Seeing the word "Propagator" is more confusing (at least to me).  

Regarding your comments about separators in the various elements that make up the panel, I will look at that and see how it might be made to work.

joetopshot:
The last commit included the cloud color/shape property changes.  I forgot to add a comment to that effect.

freeplane/src/main/java/org/freeplane/features/styles/mindmapmode/styleeditorpanel/StyleEditorPanel.java
```diff
@@ -0,0 +1,428 @@
...
+					init();
+				}
+			}
+		});
+	}
+
+	private List<IPropertyControl> getControls() {
+		final List<IPropertyControl> controls = new ArrayList<IPropertyControl>();
+		
+		for (int i=0; i<controlGroups.length; i++) {
```
dpolivaev:
```java
for (ControlGroup controlGroup : controlGroups) {
  controlGroup.addControlGroup(controls);
}
```

PR #74, opened 2017-04-30
Some documentation to explain the "Access Restriction" errors generated by Eclipse 

joetopshot:
Added comments to BigBufferedImage.java related to the use of Directbuffer.


joetopshot:
Not sure why I'm showing 7 commits instead of the one acb606e.  I think I should have done the merge and committed it before making my change.

freeplane/src/main/java/org/freeplane/core/ui/image/BigBufferedImage.java
```diff
@@ -192,6 +192,18 @@ public void run() {
 			}.start();
 		}
 
+		/**
+		 * There appears to be a bug in the Java code associated with MappedByteBuffer instances (maybe other
+		 * related classes as well?) in that closing them generates a FileNotFoundException.  For details about
+		 * attempts to get around this problem, see the following
```
dpolivaev: Our problem is rather that file.delete() is ignored and the files can not be deleted even on exit.

joetopshot:
Hopefully this is better.

PR #84, opened 2017-10-09
Replace 1.8 API call with 1.7 equivalent code.

inglor:
The java.util.Map.remove(Object key, Object value) is > 1.8 and since
the project should work with 1.7 replace this with the expanded
implementation.

Signed-off-by: Leonidas Spyropoulos <artafinde@gmail.com>


dpolivaev:
Thanks, I integrated the proposed changes

PR #87, opened 2017-11-29
User icons sorted alphabetically while loading

ziembla:
Custom icon list is explicitly sorted because on ext4 filesystem (linux) the file order is extremely messy.


dpolivaev:
Thanks for pull request and the explanation.

Issue #95, opened 2018-02-08
Create Linux desktop file


freeman42x:
This would improve user experience - easier to discover in one of the Linux menus.

Specification for the format here: https://standards.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html

Examples:

> [Desktop Entry]
> Name=Atom
> Comment=A hackable text editor for the 21st Century.
> GenericName=Text Editor
> Exec=/nix/store/6n2vw4wygn6dxx7vf35dsqg0whgk9rwh-atom-1.23.3/bin/atom %F
> Icon=atom
> Type=Application
> StartupNotify=true
> Categories=GNOME;GTK;Utility;TextEditor;Development;
> MimeType=text/plain;

> [Desktop Entry]
> Type=Application
> Exec=firefox %U
> Icon=firefox
> Comment=
> Terminal=false
> Name=Firefox
> GenericName=Web Browser
> MimeType=text/html;text/xml;application/xhtml+xml;application/vnd.mozilla.xul+xml;x-scheme-handler/http;x-scheme-handler/https;x-scheme-handler/ftp
> Categories=Application;Network;WebBrowser;

The right place to put the file would probably be under `/freeplane/resources/linux/freeplane.desktop`
I will create it and send a pull request if everything sounds ok. After that, I'll test that it works under NixOS.

htgoebel:
Some distributions already ship a desktop file, e.g. http://svnweb.mageia.org/packages/cauldron/freeplane/releases/1.3.15/3.mga6/SOURCES/freeplane.desktop?view=markup

inglor:
Also in [AUR] package which I maintain. If upstream provides one it it would be best.

[AUR]: https://aur.archlinux.org/cgit/aur.git/plain/freeplane.desktop?h=freeplane

freeman42x:
Can I add a .desktop file to this repository? That is what makes sense the most.

freeman42x:
@htgoebel @inglor Found another example, will do a merge of them:

> [Desktop Entry]
> Version=1.0
> Name=Freeplane
> Exec=/usr/bin/freeplane %F
> Terminal=false
> Icon=freeplane
> Type=Application
> MimeType=application/x-freeplane;
> Categories=Office;
> GenericName=Freeplane
> Comment=A free tool to structure and organise your information with mind mapping
> Keywords=Mindmaps; Knowledge management; Organize information; Brainstorming; ...;
> 
> [Desktop Entry]
[...]

PR #96, opened 2018-02-09
Added Linux Desktop Entry

freeman42x:
Solution for: https://github.com/freeplane/freeplane/issues/95

### DIFF
--- a/freeplane/resources/linux/freeplane.desktop
+++ b/freeplane/resources/linux/freeplane.desktop
@@ -1,0 +1,25 @@
+[Desktop Entry]
+Name=Freeplane
+GenericName=Freeplane
+GenericName[ru]=Интеллект-карты
+GenericName[de]=Mind Mapping
+GenericName[ca]=Mapa mental
+GenericName[cs]=Tvorba myšlenkových map
+GenericName[es]=Mapa de Mente
+GenericName[eu]=Gogoeta-mapak
+GenericName[fr]=Cartographie d'idées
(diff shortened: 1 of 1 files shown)


dpolivaev:
How can I test it under Ubuntu?

freeman42x:
@dpolivaev I tested it under NixOS.

To test it right now is a bit complicated since it has not been packaged yet completely as the package is relying on this .desktop changes also. If you really want to test it you could install Nix package manager under Ubuntu https://nixos.org/nix/ , https://nixos.org/nix/download.html and then
`git clone git@github.com:razvan-panda/nixpkgs.git` and checkout the `freeplane` branch. While being in the cloned repository root run `nix-env -f . -iA freeplane`. This should install freeplane and it should be available to launch via terminal or from the Ubuntu menu.

I'll try to see if I don't have any Ubuntu VM to test on it also.

dpolivaev:
All right. I shall merge it now and I shall move the file to another directory at the weekend.
I would like to change the top level directory structure a bit: currently your file would better fit into directory `freeplane/freeplane_framework/linux-resources` , but I think it would be even better to create a new directory `freeplane/freeplane-resources` and to move some of stuff under `freeplane/freeplane_framework` there. Please look into current freeplane/freeplane_framework and comment what you think about it.

freeman42x:
I had a look but it is not clear to me where it should go. I placed it under freeplane/resources since that seemed a good place to place it. If you change its location I'll need to update the Nix expression after that. That's fine as long as you can structure the project better.

PR #111, opened 2018-10-29
Enable listings environment in beamer export template

jose1711:


dpolivaev:
@fnatter could you please review this PR

igor-go:
As listings are to format code, this could be done just by invoking this environment from the written text. Do you think that we should create a kind of node attribute to enable it directly?,

I'll add the listings package load in the next version's header.

jose1711:
i don't think a separate node type is necessary. anyway the curly brackets had also to be removed - otherwise the pdf could not by typeset.

fnatter:
hello Igor,

could you please make an upstream release for inclusion in Freeplane 1.7.x?

Thanks and Best Regards!
Felix

PR #115, opened 2018-11-23
Replace deprecated methods with suggestions

inglor:
Signed-off-by: Leonidas Spyropoulos <artafinde@gmail.com>


dpolivaev (commented):
Freeplane consistently uses deprecated File.toUrl() method because toURI() incodes some characters (e.g. space characters) and I do not know if the new version still works as expected or not.

freeplane_plugin_openmaps/src/main/java/org/freeplane/plugin/openmaps/mapelements/OpenMapsViewer.java
```diff
@@ -14,7 +14,7 @@
 	private static final int WIDTH = 800;
 
 	public OpenMapsViewer () {
-		 super(new MemoryTileCache(), 4);
+		 super(new MemoryTileCache());
```
dpolivaev, line 17: I changed it as you propose here

PR #117, opened 2018-12-19
Fix for #2386 Special characters in add-on translations wrongly displayed

HaaiHenkie:
Added method `escapeUtf8` to `org.freeplane.core.util.TextUtils`. Method
`escapeUtf8` is a stripped down version of `StringEscapeUtils.escapeJava`. It
only escapes UTF-8 non-ASCII characters. It escapes them to \uXXXX where
XXXX is the hex code of the escaped character. It solves this bug
because it does not escape backslashes and double quotes.

Replaced method `StringEscapeUtils.escapeJava` with
`TextUtils.escapeUtf8` on line 456 of
`org.freeplane.main.addons.AddOnProperties`.

As part of the fix of this bug the add-on `devtools` is also changed to
use `TextUtils.escapeUtf8`.

### DIFF
--- a/freeplane/src/main/java/org/freeplane/core/util/TextUtils.java
+++ b/freeplane/src/main/java/org/freeplane/core/util/TextUtils.java
@@ -1,8 +1,12 @@
 package org.freeplane.core.util;
 
+import java.io.IOException;
+import java.io.StringWriter;
+import java.io.Writer;
 import java.text.DecimalFormat;
 import java.text.MessageFormat;
 import java.text.SimpleDateFormat;
+import java.util.Locale;
 import java.util.regex.Pattern;
 
 import org.freeplane.core.resources.ResourceBundles;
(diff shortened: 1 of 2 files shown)


vboerchers:
That's great, Henk!

Just one thing: Could you please move the reference to Commons Lang from normal comment into JavaDoc of  `TextUtils.escapeUtf8` and remove the JavaDoc for the new private `TextUtils` methods? We need documentation of the API but internal comments [shouldn't be necessary](https://refactoring.guru/smells/comments).

I also have a question: What about translations spanning over multiple lines? Are those handled correctly? If not we might need an additional `replace("\n", "\\n")` or `"\u000A"`. (That could be considered after merging both PRs.)

Best regards
Volker

HaaiHenkie:
Hi Volker,

I adapted the JavaDoc of `TextUtils.escapeUtf8`.

Is it possible to do multi-line editing in attributes? I just tried but was not able to do that. Alt+Enter while editing a attribute brings up the SimplyHTML editor for the node text. So I think an add-on developer needs to put '\n' manually anyway for multi-line strings. How many add-on developers need to translate multi-line strings anyway. I am probably one of the few. 

Since I have over 150 strings maintenance and translation is not doable by editing attributes. I do all translations now in ResourceBundle Editor (eclipse plugin, see screenshot below). It does all escaping of end of line and non-ASCII characters automatically. I have a small node based script to import the translation properties files into the translation attributes.

Best regards,

Henk

image

PR #125, opened 2019-01-20
Fix for Javadoc escapeUtf8 in org.freeplane.core.util.TextUtils

HaaiHenkie:
In merged pull request #117 the new method `escapeUtf8` was added to
`org.freeplane.core.util.TextUtils`. There is a small mistake in the
example in Javadoc. It reads:

> input string: jalapeño
> output string: jalapeño

It should read

> input string: jalapeño
> output string: jalape\u00F1o

Somehow `\u00F1` gets unescaped when creating the Javadoc. This is
strange because the Javadoc preview shows it correct A double backslash
in the same text is not escaped, but it is enclosed with `<code>` tags.
I have enclosed `\u00F1` also with `<code>` tags. Hopefully that will
solve the problem.

image

### DIFF
--- a/freeplane/src/main/java/org/freeplane/core/util/TextUtils.java
+++ b/freeplane/src/main/java/org/freeplane/core/util/TextUtils.java
@@ -212,7 +212,7 @@ public class TextUtils {
 	 * Example:
 	 * <pre>
 	 * input string: jalapeño
-	 * output string: jalape\u00F1o
+	 * output string: jalape<code>\u00F1</code>o
 	 * </pre>
 	 *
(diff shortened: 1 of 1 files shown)


HaaiHenkie:
I did a `gradle clean build` and unfortunately this did not fix the problem

dpolivaev:
You can fix it as described at https://stackoverflow.com/questions/32020170/escape-unicode-escaping-in-java

dpolivaev:
Fixed

HaaiHenkie:
Actually, while I was at the supermarket I thought of exactly the same solution

PR #157, opened 2020-04-12
Contributing Appstream metadata[1] for Freeplane.

olea:
[1] https://www.freedesktop.org/software/appstream/docs/index.html

This is used by modern distros and Flathub.


fnatter:
@dpolivaev : This is for flatpaks, so it does not affect our upstream .deb build. It makes sense to have it included.

fnatter:
@olea: Thanks for the contribution!

olea:
> @dpolivaev : This is for flatpaks, so it does not affect our upstream .deb build. It makes sense to have it included.

The Appstream metadata is currently used by modern Linux desktops (GNOME, KDE...) so it's highly recommended to be included in your Linux packages. As you said for Flatpak/Flathub is a hard requirement.

Thanks for accepting :-)

PR #193, opened 2021-04-19
User/viaa

viaa:
The small change (1 line added) to SetLinkByTextFieldAction.java is just to be able to paste in links, paths that contain double quotes, like "c:\temp" instead of c:\temp.
The double-quotes come from Windows Explorer, doing a right-click on a file while holding the shift key will show the contextual menu in Windows Explorer and (the shift key) will add "Copy as path" option. This copies the path of the file but adds double quotes.


dpolivaev:
I applied the idea

PR #198, opened 2021-05-17
Move a node to top, bottom margin or a corner

firvorski:
Hi,

this is a pull request for the change to enable quick positioning of nodes at the top or the bottom of the screen or in a corner. 

One thing I didn't know how to go about is all the translations so that is missing.

Also, I wanted to create a submenu so that there aren't too many entries there, but I can do that if it gets included.


dpolivaev:
Thanks. Yes, I would like to merge it, after a submenu is created and all menu items are properly sorted.
How do you want to call the submenu and where do you want to have it?
I was thinking about calling it "Scrolling" and placing it under "View"

freeplane/src/main/java/org/freeplane/features/map/IMapSelection.java
```diff
@@ -32,7 +32,8 @@
 public interface IMapSelection {
 
 	public enum NodePosition {
-		WEST, CENTER, EAST;
+		WEST, CENTER, EAST, NORTH, SOUTH,
```
dpolivaev, line 35: What do you think about renaming the enumeration items to LEFT, BOTTOM_LEFT etc?

dpolivaev:
I implemented the changes I suggested and merged your pull request.

dpolivaev:
The feature is available in preview 1.9.3_01

firvorski:
Awesome! Thanks a lot!
I didn't expect the turnaround time to be so swift, I would have done what
you suggested but was too slow to react, will keep it in mind for the next
time though.

dpolivaev:
That's fine, I like that we could do it together.
I mostly spend time on Freeplane at the weekend and I am usually not active during the rest of the week.

PR #200, opened 2021-05-21
Bugfix/malformed translation val zh_CN

danix800:
Screenshot from 2021-05-21 14-42-39


danix800:
After fix: 

Screenshot from 2021-05-22 01-03-39

PR #202, opened 2021-05-24
emulate vim-like simple keyboard navigation

danix800:
Inspired by:
https://sourceforge.net/p/freeplane/discussion/758437/thread/e904441d/?limit=25

Currently supported key mapping in this mode:

1. 'h', 'j', 'k', 'l' for node selection
2. 'H', 'J', 'K', 'L' for map scrolling
3. '+', '-' for zooming
4. 'i' for edit current
5. 'n', 'N' for next/prev search
6. ':' for command search

Not supported:

1. 'o', 'O' for new node, we have 'Enter', 'Insert'
2. 'ESC' for NORMAL MODE, don't need this
3. '/' for search, we have ctrl+F

Other useful mapping could be added in the future.


danix800:
This is still a navigational related issue I tried to solve.

As a big vim fan I personally like this feature but surely it needs further discussion from the whole picture, which I didn't grasp for now.

When I tried to read the document map (`F1`) this weekend at home I was on my laptop without mouse, I used track-point and arrow keys, and I turned the key type action from default `EDIT_CURRENT` to `IGNORE` to avoid inadvertent key strokes. Soon after I realized maybe I can turn my keystrokes into vim mode, at least in navigational point of view.

I use vim like editing for almost everything I can get, so this is just out of habit, maybe too much personal stuff? That's what I'm worried about. I searched google for this but not too much discussions about this.

Hope to see your responds.

danix800:
Closed for now, I need to do more investigations.

PR #203, opened 2021-05-25
set map overview cursor to CROSSHAIR_CURSOR

danix800:
Use `old_cursor` seems to be redundant, `Cursor.getDefaultCursor()` can do the work, so I removed this variable.


danix800:
For preview: 

Screenshot from 2021-05-25 00-54-34

PR #207, opened 2021-05-26
support overview map move & resize

danix800:
1. default to 320x320 anchor to SE; bounds persisted
2. resizable (border center points / corner points);
3. movable (other places of the border);
4. switch anchor point (attach point) by double clicking corner points. (tagged by an anchor emoji in the corresponding corner)

Screenshot from 2021-05-27 02-09-31


danix800:
Hi, @dpolivaev,  @ldmpub, when you have time please take a look at this PR, thanks.

dpolivaev:
Sure, at the weekend. Thanks.

dpolivaev:
I have too many questions.
Could we have a session e.g. in zoom to discuss them?

dpolivaev:
On the one hand, some parts of the implementation do not work,
and on the other hand, we need to fix names in the code so that I can at least understand the intention.
I would be very grateful if you could find time for a session.

dpolivaev:
And after I understand the intention we need to discuss whether there we can simplify the logic.

freeplane/src/main/java/org/freeplane/view/swing/map/MapViewPane.java
```diff
@@ -206,46 +232,383 @@ public void mouseDragged(MouseEvent e) {
...
+            }
+        };
+
+        private final int cursor;
+
+        Locations(int cursor) {
+            this.cursor = cursor;
+        }
+
+        abstract Point getPoint(Rectangle r);
```
dpolivaev: I do not understand what is the meaning of the point.
Which point is it?
Could you rename the method so that it better reveals the intent?

freeplane/src/main/java/org/freeplane/view/swing/map/MapViewPane.java
```diff
@@ -206,46 +232,383 @@ public void mouseDragged(MouseEvent e) {
...
+        public Cursor getCursor() {
+            return Cursor.getPredefinedCursor(cursor);
+        }
+    }
+
+    private interface ResizableBorder extends Border {
+        Cursor getResizeCursor(MouseEvent e);
+    }
+
+    private class DefaultResizableBorder implements ResizableBorder, SwingConstants {
```
dpolivaev: Instead of having multiple inner classes I would rather have a new package. For extracting this class to external file it should be static. And it looks like it could be possible because it does not depend on the internal state.

Making classes static and extracting them to separate files helps to make dependencies between them explicit and to reduce coupling which means reducing of the complexity.

freeplane/src/main/java/org/freeplane/view/swing/map/MapViewPane.java
```diff
@@ -206,46 +232,383 @@ public void mouseDragged(MouseEvent e) {
...
+                return new Point(getDeltaX(deltaX, pr), getDeltaY(deltaY, pr));
+            default:
+                return new Point(deltaX, deltaY);
+            }
+        }
+    }
+
+    private final static int COMPAT_MOVE_CURSOR = Compat.isMacOsX() ? Cursor.HAND_CURSOR : Cursor.MOVE_CURSOR;
+
+    private enum Directions {
```
dpolivaev: I do not understand if there is a real difference between "Locations" and "Directions", for me it looks like the same thing because both enums have the same elements.

freeplane/src/main/java/org/freeplane/view/swing/map/MapViewPane.java
```diff
@@ -206,46 +232,383 @@ public void mouseDragged(MouseEvent e) {
...
+                rect.setLocation(loc.getPoint(r));
+                g.setColor(mapViewBackground);
+                g.fillRect(rect.x, rect.y, rect.width - 1, rect.height - 1);
+                g.setColor(complementaryColor(mapViewBackground));
+                g.drawRect(rect.x, rect.y, rect.width - 1, rect.height - 1);
+            }
+        }
+
+        @Override
+        public Cursor getResizeCursor(MouseEvent e) {
```
dpolivaev: It looks like this method always returns COMPAT_MOVE_CURSOR.
It looks like something could be wrong here

freeplane/src/main/java/org/freeplane/view/swing/map/MapViewPane.java
```diff
@@ -292,14 +643,97 @@ public void componentResized(ComponentEvent e) {
         });
     }
 
+    private MapOverviewAttachPoint getMapOverviewAttachPoint() {
```
dpolivaev: Could a double click on the anchor be used to reset the position?

dpolivaev (commented):
Let us talk about all things in Zoom when you find time

freeplane/src/main/java/org/freeplane/view/swing/map/MapViewPane.java
```diff
@@ -1,613 +1,38 @@
...
 import org.freeplane.core.resources.IFreeplanePropertyListener;
 import org.freeplane.core.resources.ResourceController;
-import org.freeplane.core.ui.components.resizer.UIComponentVisibilityDispatcher;
-import org.freeplane.core.util.Compat;
 import org.freeplane.features.map.IMapChangeListener;
 import org.freeplane.features.map.MapChangeEvent;
 import org.freeplane.features.mode.Controller;
-import org.freeplane.features.ui.IMapViewManager;
 import org.freeplane.features.ui.ViewController;
+import org.freeplane.view.swing.map.overview.MapOverviewImage;
```
dpolivaev: Could you move also MapViewPane to the new package so that most classes of the new package do not need to be public, visible, and known outside the package?
By the way, in the best of all possible worlds, the first commit would only extract the classes and the second only move all classes to the new package, it was nice but it is not really important.

freeplane/src/main/java/org/freeplane/view/swing/map/overview/MapOverviewUtils.java
```diff
@@ -0,0 +1,100 @@
...
+import java.awt.Insets;
+import java.awt.Rectangle;
+
+import javax.swing.JScrollBar;
+import javax.swing.JScrollPane;
+
+import org.freeplane.core.resources.ResourceController;
+import org.freeplane.view.swing.map.MapView;
+
+public class MapOverviewUtils {
```
dpolivaev: In recent years, I become more addicted to clean code and sensitive to ways to improve it, so I hope you allow me to share some ideas with you as I see ways to improve the code.

Having a collection of static methods inside a utility class always seems suspicious. Actually, such classes are needed only in a specific situation when you want to extend some existing library or framework in a generic way. And even in such cases it is better to create a separate extention class per extended class or even per extension. 

Actually even a word Utilities (or its abbreviation)  as a part of the class name is suspicious.

freeplane/src/main/java/org/freeplane/view/swing/map/overview/MapOverviewUtils.java
```diff
@@ -0,0 +1,100 @@
...
+    static Color complementaryColor(final Color color) {
```
dpolivaev: This method could be put in a dedicated class

freeplane/src/main/java/org/freeplane/view/swing/map/overview/MapOverviewUtils.java
```diff
@@ -0,0 +1,100 @@
...
+        String rawAttachPoint = resourceController.getProperty(MapOverviewConstants.PROP_MAP_OVERVIEW_ATTACH_POINT,
+                MapOverviewAttachPoint.SOUTH_EAST.name());
+        try {
+            return MapOverviewAttachPoint.valueOf(rawAttachPoint);
+        } catch (IllegalArgumentException e) {
+            return MapOverviewAttachPoint.SOUTH_EAST;
+        }
+    }
+
+    public static void convertOriginByAttachPoint(MapView mapView, Rectangle bounds) {
```
dpolivaev: I think part of the best practices is putting public methods ahead of the private methods they call so that when someone reads a class its purpose comes first and implementation details follow.

Because this method is used only by classes related to MapViewPane I would move it to MapViewPane and remove MapView as its parameter. 

ResizePanelMouseHandler could be given a reference to MapViewPane or obtain it from the event. It does not need to know MapView directly.

freeplane/src/main/java/org/freeplane/view/swing/map/overview/MapOverviewUtils.java
```diff
@@ -0,0 +1,100 @@
...
+            break;
+        case NORTH_EAST:
+            bounds.setLocation(right - bounds.x - vsw - 1 - bounds.width, bounds.y);
+            break;
+        case NORTH_WEST:
+            break;
+        }
+    }
+
+    public static int getBoundedValue(int value, int min, int max) {
```
dpolivaev: class MapOverviewMathOperations?

freeplane/src/main/java/org/freeplane/view/swing/map/overview/MapOverviewUtils.java
```diff
@@ -0,0 +1,100 @@
...
+        return Math.max(Math.min(value, max), min);
+    }
+
+    public static Rectangle getMapOverviewBounds(MapView mapView) {
```
dpolivaev: Same hints as for convertOriginByAttachPoint

freeplane/src/main/java/org/freeplane/view/swing/map/overview/MapOverviewUtils.java
```diff
@@ -0,0 +1,100 @@
...
+        int x = elements[0];
+        int y = elements[1];
+        int width = getBoundedValue(elements[2], MapOverviewConstants.MIN_SIZE, MapOverviewConstants.MAX_SIZE);
+        int height = getBoundedValue(elements[3], MapOverviewConstants.MIN_SIZE, MapOverviewConstants.MAX_SIZE);
+        Rectangle bounds = new Rectangle(x, y, width, height);
+        convertOriginByAttachPoint(mapView, bounds);
+        return bounds;
+    }
+
+    public static double getBestScale(Rectangle source, double scaledOverviewBoundWidth,
```
dpolivaev: I would move this method to MapOverviewImage and remove its first argument.

dpolivaev (commented):
MapOverviewUtils review.

Please let me know how useful you find it.

freeplane/src/main/java/org/freeplane/view/swing/map/overview/ResizableBorder.java
```diff
@@ -0,0 +1,146 @@
...
+import java.awt.Point;
+import java.awt.Rectangle;
+import java.awt.event.MouseEvent;
+
+import javax.swing.SwingConstants;
+import javax.swing.border.Border;
+
+import org.freeplane.view.swing.map.MapView;
+
+public class ResizableBorder implements Border, SwingConstants {
```
dpolivaev: In the best of all possible worlds, I would put ResizableBorder and ResizePanelMouseHandler in their own package because they are not specific to the map overview

dpolivaev:
I need to thank you for the efforts you already put into the refactoring.
I appreciate your work and effort. It is a real improvement.

dpolivaev:
I would like to suggest to minor changes to the functionality. Feel free to implement the suggestions if you like them.

1. I would show the zoom always and not only if the main toolbar is not visible. If you like you could also make it a separate option.
2. I would attach the overview panel to the corresponding corner on double click at the resizable border corners. Using the menu actions for this purpose seems to be less convenient for people using the mouse.

ldmpub:
+1 thanks to you two, great work all together 👍.

dpolivaev:
Please let me know when you are done so that I can merge it.

danix800:
I'm still working hard on refactoring, 

> I need to thank you for the efforts you already put into the refactoring.
> I appreciate your work and effort. It is a real improvement.

I'm still busy on both refactoring and minor visual adjustment so I might not able to respond all your questions and advice in time. But I'll read all of your suggestions carefully.

freeplane/src/main/java/org/freeplane/view/swing/map/overview/ResizableBorder.java
```diff
@@ -0,0 +1,146 @@
...
+    static final int SIZE = 6;
```
dpolivaev: I would make size a parameter, move the constant to the place where overview DEFAULT_SIZE is defined and calculate it using Quantity

freeplane/src/main/java/org/freeplane/view/swing/map/overview/ResizableBorder.java
```diff
@@ -0,0 +1,146 @@
...
+        if (! bounds.contains(pt)) {
+            return Cursor.getDefaultCursor();
+        }
+
+        Rectangle contentAreaBounds = new Rectangle(SIZE, SIZE, width - 2 * SIZE, height - 2 * SIZE);
+        if (contentAreaBounds.contains(pt)) {
+            return Cursor.getDefaultCursor();
+        }
+
+        Rectangle actionableRegion = new Rectangle(SIZE, SIZE);
```
dpolivaev: What do you think about increasing the size of this region, probably doubling it, so that it is easier to find it with the mouse cursor?

freeplane/src/main/java/org/freeplane/view/swing/map/overview/MapOverviewConstants.java
```diff
@@ -0,0 +1,21 @@
+package org.freeplane.view.swing.map.overview;
+
+import java.awt.Color;
+
+import org.freeplane.api.LengthUnit;
+import org.freeplane.api.Quantity;
+
+public class MapOverviewConstants {
```
dpolivaev: Because of the reasons also valid for the Utils, I would move the constants to the classes they are related to. Introducing dedicated classes for constant values if rather needed of it allows you to hide some classes inside its package and keep the public API small. (By the way, interfaces are better for it because there you nobody expects behavior)

dpolivaev (commented):
Some other points

freeplane/src/main/java/org/freeplane/view/swing/map/overview/MapOverviewConstants.java
```diff
@@ -0,0 +1,21 @@
...
+    public final static String PROP_MAP_OVERVIEW_VISIBLE = "mapOverviewVisible";
+    public final static String PROP_MAP_OVERVIEW_VISIBLE_FS = "mapOverviewVisible.fullscreen";
+
+    public final static String PROP_MAP_OVERVIEW_PREFIX = "map_overview_";
+    public final static String PROP_MAP_OVERVIEW_ATTACH_POINT = PROP_MAP_OVERVIEW_PREFIX + "attach_point";
+    public final static String PROP_MAP_OVERVIEW_BOUNDS = PROP_MAP_OVERVIEW_PREFIX + "bounds";
+
+    public final static int DEFAULT_SIZE = (int) new Quantity<LengthUnit>(240, LengthUnit.pt).in(LengthUnit.px).value;
```
dpolivaev: instead of using `(int) ... .in(LengthUnit.px).value` I would rather use `.toBaseUnitsRounded()` because it is exactly what this method should do.

It does not contain word "pixels" because Quantity is not necessarily a length.

freeplane/src/main/java/org/freeplane/view/swing/map/overview/ResizableBorder.java
danix800: Already considered, the actionable area is expanded. Thick segments is for resizing, thin line for move.

Screenshot from 2021-05-30 18-34-20

danix800:
> I would like to suggest to minor changes to the functionality. Feel free to implement the suggestions if you like them.
> 
> 1. I would show the zoom always and not only if the main toolbar is not visible. If you like you could also make it a separate option.
> 2. I would attach the overview panel to the corresponding corner on double click at the resizable border corners. Using the menu actions for this purpose seems to be less convenient for people using the mouse.

danix800:
> I would like to suggest to minor changes to the functionality. Feel free to implement the suggestions if you like them.
> 
> 1. I would show the zoom always and not only if the main toolbar is not visible. If you like you could also make it a separate option.
> 2. I would attach the overview panel to the corresponding corner on double click at the resizable border corners. Using the menu actions for this purpose seems to be less convenient for people using the mouse.

1. Reasonable, I'll remove the toolbar visibility checking.
2. I also found that left double click will be useful (menu actions will also be retained for heavy keyboard users).

freeplane/src/main/java/org/freeplane/view/swing/map/MapViewPane.java
```diff
@@ -1,251 +1,39 @@
...
-            }
-            return image;
-        }		
-		
-
-		@Override
-        protected void paintComponent(Graphics g) {
-			final Graphics2D g2 = (Graphics2D)g;
-			final AffineTransform transform = g2.getTransform();
-			final double scaleX = transform.getScaleX();
```
danix800, line 103: I want to know when or under which environments will the scaleX NOT equal to 1? could you give some examples?

dpolivaev: On MacOS when 4K monitor is used with resolution 1080 scaleX=scaleY=2

danix800: :ok_hand: Thanks!

freeplane/src/main/java/org/freeplane/view/swing/map/MapViewPane.java
danix800: Not reproducible for me, need more investigation.

freeplane/src/main/java/org/freeplane/view/swing/map/MapViewPane.java
danix800: Will be available in next commits.

freeplane/src/main/java/org/freeplane/view/swing/map/MapViewPane.java
danix800: Will be clarified in next commits.

freeplane/src/main/java/org/freeplane/view/swing/map/MapViewPane.java
danix800: Will be treated in next commits.

freeplane/src/main/java/org/freeplane/view/swing/map/MapViewPane.java
danix800: Will be clarified in next commits.

freeplane/src/main/java/org/freeplane/view/swing/map/MapViewPane.java
danix800: Will be treated in next commits.

freeplane/src/main/java/org/freeplane/view/swing/map/MapViewPane.java
dpolivaev: The latest version worked fine on my machine

freeplane/src/main/java/org/freeplane/view/swing/map/overview/MapOverviewConstants.java
danix800: Will be treated in next commits.

freeplane/src/main/java/org/freeplane/view/swing/map/overview/ResizableBorder.java
danix800: Will be treated in next commits.

freeplane/src/main/java/org/freeplane/view/swing/map/overview/MapOverviewUtils.java
danix800: Already moved into MapOverviewImage, but could you reconsider removing the first parameter, I might want to keep it since mapView.getInnerBounds() seems to be none trivial, so I'd calculate only once and pass it down.

I can understand that from clean code perspective too much parameters are bad so if possible I'll try to reduce number of parameters of method.

freeplane/src/main/java/org/freeplane/view/swing/map/overview/MapOverviewConstants.java
danix800: Will be treated in next commits.

freeplane/src/main/java/org/freeplane/view/swing/map/overview/ResizableBorder.java
danix800: Will be treated in next commits.

danix800:
Please give a look at the latest commit, there might still be issues but I'll come back tomorrow! Thanks.

freeplane/src/main/java/org/freeplane/view/swing/map/overview/MapOverviewUtils.java
dpolivaev: Reducing the number of parameters usually leads either to either moving logic to the data or to creating new abstractions.
I see your argument that you want to avoid expansive calculation of the map view inner bounds.
Probably MapView could cache the results until they are invalidated and remove the cache when MapView. paintChildren(final Graphics g) is called. But it is only something to be considered, as freeplane code is not clean enough to have all good principles satisfied. So I think we can leave it as it is now.

dpolivaev:
I like the improvements. I hope, you too, don't you?

The new thick lines at the corners and at the middle points of the border appear too fat. For me they look distractive and a bit ugly. You could make them thinner and use another stroke (dash vs solid) so they they look different from the other segments.

You can do not need to fill all pixels parts of the border and let some space around the overview not filled so that the mouse cursor changes but the map remains visible.

dpolivaev:
Hello Ding, because your feature branch is living more than 1-2 days. To integrate earliy it would be nice if you could regularly merge the development branch 1.9.x from the freeplane repository into your branch. For doing so you can add the second remote. Probably you know how it works, but just in case it is useful, here it is.

```
git remote add freeplane https://github.com/freeplane/freeplane.git
git merge freeplane/1.9.x
```

danix800:
> Hello Ding, because your feature branch is living more than 1-2 days. To integrate earliy it would be nice if you could regularly merge the development branch 1.9.x from the freeplane repository into your branch. For doing so you can add the second remote. Probably you know how it works, but just in case it is useful, here it is.
> 
> ```
> git remote add freeplane https://github.com/freeplane/freeplane.git
> git merge freeplane/1.9.x
> ```

I'll locally rebase to 1.9.x after the final commit is ready, there won't be integration problems.

dpolivaev:
Generally, I recommend merging early and checking for the integration problems early.
If it works it is fine, and if it does not work you become aware of it earlier and have less to fix.

Merging early is useful in many cases. It is particularly important if you refactor because after refactoring merge conflicts are more likely. The more developers work on a project, the more important frequent merges become.

dpolivaev:
If you see any merge conflicts you can not resolve easily it means that we need to do it together. You can always revert the merge in this case.

dpolivaev:
On my computer, I constantly locally merge 1.9.x into your branch when I pull your changes, and I have not seen any problems so far. And actually, I think that you could even merge from freeplane/1.9.x before any push.

dpolivaev:
I consider it to be the best practice, and we can discuss all pros and cons if you like.

danix800:
Right, thanks for remind. I'm keeping track of the upstream so that my branch won't go wild.

dpolivaev:
my-local-merges

dpolivaev:
I checked your latest version. It looks nice.
I have seen that if I double-click at the corner not only the position but also the dimension is reset.
I would rather like it if these operations were independent.

dpolivaev:
And I need to concentrate on my work now. I come back to the issue on the day I am free.

EdoFro:
Dear Dimitry, dear Ding Fei,
I have been following this commit and your conversation and I wanted to say thank you for  all your efforts and the great work, and to tell you that your conversation has been very interesting and I have learned a lot on collaborative work, good practices and so on. 
Maybe it could be a good idea, after the final merge, to point a link to this commit from the Freeplane wiki as an example on how to collaborate with the project. 
I know that making a ‘How to’ document explaining the procedure is complicated and time consuming, that’s why I think this merge can become a nice ‘how to’ example. 
jm2c
BR
edo

EdoFro:
Ah! and please do the final merge including all single commits. I know there is an option to merge all commits into a single one, but then we will miss all the history and all the explanation what each good practice do in the code in the practice. 
thanks again

dpolivaev:
@EdoFro I enjoy working together with @danix800  a lot too.
I don't think we can convert this thread into a how-to document because every person is different and all we need is openness, goodwill, and communication.

danix800:
> @EdoFro I enjoy working together with @danix800 a lot too.
> I don't think we can convert this thread into a how-to document because every person is different and all we need is openness, goodwill, and communication.

I'm enjoying this pleasant journey too! @dpolivaev is of great help to this PR and without him I'm afraid I cannot make this feature reaching its current state. He's patient, generous and open to share his knowledge with me and I've learned a lot.

The other important thing I've learned is to stay open to communicate. Even I'm a bad English speaker (sorry for not warning you in advance :smile: ) @dpolivaev encouraged and persuaded me to have video sessions when needed. With his questions, he actually taught and pushed me to reflectively think about some of the original code and redefine the feature in a more clear way. With these premises setup then the implementation becomes quite easier than I thought at first.

I really want to say thank you again @dpolivaev.

danix800:
> I checked your latest version. It looks nice.
> I have seen that if I double-click at the corner not only the position but also the dimension is reset.
> I would rather like it if these operations were independent.

The double click behavior is reworked, please check the latest commit and if there's no other issues then this can be merged. It's really cool we've reached here! Thanks!

danix800:
> Ah! and please do the final merge including all single commits. I know there is an option to merge all commits into a single one, but then we will miss all the history and all the explanation what each good practice do in the code in the practice.
> thanks again

It's not necessary to preserve the feature branch history if upstream prefers, especially when the history is not that neat (e.g. https://github.com/freeplane/freeplane/pull/207/commits/ee14cbefeaaa601795d065bb2bda9f43d50f6d52 and https://github.com/freeplane/freeplane/pull/207/commits/ac2ee8e0e3a947e0089339f668a68be87df3e533 not clearly describing what they did). Squash them into one commit is OK for me. I think the whole conversations here won't be lost because of the squash.

ldmpub:
I'm glad to read all these "good vibrations" 👍.

dpolivaev:
I found and fixed 2 bugs in https://github.com/freeplane/freeplane/commit/78a2db3a6a31a619af6b66c8ecab02509c57a94c

- The property should be always synchronised with the visibility
- `mapView.getParent().getParent()` does not necessarily returns scroll panes. it depends on look-and-feel

dpolivaev:
I am going to release 1.9.3 before these improvements

ldmpub:
I'm eager to test this new feature improved!

danix800:
> I found and fixed 2 bugs in [78a2db3](https://github.com/freeplane/freeplane/commit/78a2db3a6a31a619af6b66c8ecab02509c57a94c)
> 
> * The property should be always synchronised with the visibility
> * `mapView.getParent().getParent()` does not necessarily returns scroll panes. it depends on look-and-feel

Could you think about the property and visibility synchronization fix again? This menu item is for hiding the overview for currently open mapview, as [this](https://sourceforge.net/p/freeplane/discussion/758437/thread/494dfaa629/?page=1&limit=25#073f) discussion requested for the controlling of overview for multiple map views separately.

Maybe the naming `Hide` is ambiguity. If you agree with the designing, I'm going to revert this logic back and renaming `Hide` to `Hide current overview` or something alike.

danix800:
> I am going to release 1.9.3 before these improvements

Really glad to see this!

danix800:
> > I found and fixed 2 bugs in [78a2db3](https://github.com/freeplane/freeplane/commit/78a2db3a6a31a619af6b66c8ecab02509c57a94c)
> > 
> > * The property should be always synchronised with the visibility
> > * `mapView.getParent().getParent()` does not necessarily returns scroll panes. it depends on look-and-feel
> 
> Could you think about the property and visibility synchronization fix again? This menu item is for hiding the overview for currently open mapview, as [this](https://sourceforge.net/p/freeplane/discussion/758437/thread/494dfaa629/?page=1&limit=25#073f) discussion requested for the controlling of overview for multiple map views separately.
> 
> Maybe the naming `Hide` is ambiguity. If you agree with the designing, I'm going to revert this logic back and renaming `Hide` to `Hide current overview` or something alike.

Or how about keeping `Hide` for property and adding another one for `Hide current overview`?

dpolivaev:
If I click on "hide" the way it was and then looked at the map pop-up menu, the entry "map overview" was selected although the overview was hidden. It felt buggy and at least not consistent.

Certainly you can consider saving if map overview is shown per map instead of as user setting. However I do not see why this control should be treated differently from the others and why this setting should depend on the map. 

To be honest, I often think that the users bring their suggestions not because they really need them but just because they believe it would make the app better for somebody else. In my eyes increasing complexity makes the app rather worse .

How do you see it?

dpolivaev:
Also, if you hide overview only for current map, how can you make it visible again?

danix800:
> Also, if you hide overview only for current map, how can you make it visible again?

I agree with simplicity.

ldmpub:
Hi,

If I can give you my opinion on these points from an "advanced" user point of view 😉

> you can consider saving if map overview is shown per map instead of as user setting. However I do not see why this control should be treated differently from the others and why this setting should depend on the map.

I feel like an "app session" setting is perfect for this hide/show map overview feature.

> To be honest, I often think that the users bring their suggestions not because they really need them but just because they believe it would make the app better for somebody else.

I clearly do not personally agree 😄. When, as a user, I suggest a change ... it's only for my "own" benefits 😊.

dpolivaev:
> I clearly do not personally agree 😄. When, as a user, I suggest a change ... it's only for my "own" benefits 😊.

Well, then it should be easy for you to explain the situation when you need one or another feature.
In many cases the needs do not become clear enough to me.

dpolivaev:
In this particular case the question was about why should user be able to hide map overview only for one specific map view.
I don't know.

EdoFro:
Just my humble opinion:
I'm also for simplicity, and I think You are free to implement this feature as you like, because as more complex it gets more bugs possibilities it has, and the most probable is that you will be the ones that will have to correct those bugs later. So, please don't take all user ideas so seriously. It's good to listen, but the decisions are up to you. We, the users, don't know how complex it is to implement and mantain a new feature. We are free to propose new features, because You are free to implement them or not.

That beeing said:
I think the motivation for the user to ask for this was that he works with multiple maps visible at the same time and he has screen space limitations. I want to clarify that I work differently and I haven't have this problem yet. I just open or close the **Map Overview** and continue working.

Just my thoughts:
I agree that the **Map Overview** does not depend on the map itself
But It may depend on the map **view**.

Currently **Toolpanel**, **Filterbar**, and others UI elements don't depend on the map or the view, but the **Outline View** does depend on the view only: You can have the same map in two different views and **in one** you can see the tipical mindmap layout and in **the other**  in outline view mode. Maybe **Map Overview** can have a similar logic.

My feature request (if posible):
When **Map Overview** visible, if the user selects a **node** that is **under** the **Map Overview window** then the **map should scroll**. The same way as the currently logic when selecting a node that is beyond the screen.
Sometimes, when working keyboard only, I add new nodes and they appear behind the **Map Overview** and I can't see what I'm editing.
It is not so important, because the shortcut to hide the **Map Overview** works while editing a node too and it hides it and the node doesn't loose focus, so I can continue with the editing without major issues. But it still would be nice that the map would scroll automatically.

BR
edo

dpolivaev:
I played with this idea and it looks feasible and doable.
But I could not finish, I need Ding Fei for it because I do not understand some details of the implemented layout algorithm.

@danix800 what do you think about 2- hours of pair programming to get it done? I think the challenge is hard enough to join our efforts on it.

we can do it as we both find time, no hurry.

dpolivaev:
@EdoFro @danix800 it took me almost the whole day to finish it.
https://github.com/freeplane/freeplane/commit/5ee463aa6d4b09c16cc5a3ee7024ef031138bd2b "Adjust scrolling to move selected node outside of the map overview"

PR #209, opened 2021-05-27
remove default colliding accelerator key

danix800:
removed 'UseplainTextAction', colliding with 'ShowFormatPanel'


dpolivaev:
I checked which action was actually executed and kept it. So I kept another one so that the visible behavior does not change.

danix800:
:ok_hand:

PR #211, opened 2021-05-27
add map free scroll / node selection actions

danix800:
1. Scroll left/up/right/down in fixed steps;
2. Select left/up/right/down like arrow keys.


dpolivaev:
Currently, if you execute "scroll up" the map goes down and on "scroll left" it goes right.
What do you think about reversing actions "scroll left" and "scroll right" as well as "scroll up" and "scroll down"?

danix800:
> Currently, if you execute "scroll up" the map goes down and on "scroll left" it goes right.
> What do you think about reversing actions "scroll left" and "scroll right" as well as "scroll up" and "scroll down"?

Sure, I've changed scrolling directions.

PR #215, opened 2021-06-03
add action for restart

danix800:
Inspired by eclipse's restart menu item. When run in normal (via scripts) mode, the scripts serve as the launcher in eclipse's term. 

But when run/debug in eclipse there's no common launcher delegate for eclipse. I use eclipse so I'll see if there's any easy way to workaround this. Other IDEs may or may not be able to support this kind of conditional relaunch. For gradle run & debug the build script can be amended to work in this way but I'm not familiar with gradle so this will be postponed too.


danix800:
With this [JDT](https://bugs.eclipse.org/bugs/show_bug.cgi?id=574034) request and

```
diff --git a/freeplane_debughelper/eclipse/freeplane.launch b/freeplane_debughelper/eclipse/freeplane.launch
index 3b6808225..908006b10 100644
--- a/freeplane_debughelper/eclipse/freeplane.launch
+++ b/freeplane_debughelper/eclipse/freeplane.launch
@@ -12,6 +12,8 @@
     <stringAttribute key="org.eclipse.jdt.launching.MAIN_TYPE" value="org.freeplane.launcher.Launcher"/>
     <stringAttribute key="org.eclipse.jdt.launching.MODULE_NAME" value="freeplane_debughelper"/>
     <stringAttribute key="org.eclipse.jdt.launching.PROJECT_ATTR" value="freeplane_debughelper"/>
+    <intAttribute key="org.eclipse.jdt.launching.RELAUNCH_EXIT_CODE" value="194"/>
+    <booleanAttribute key="org.eclipse.jdt.launching.RELAUNCH_ON_EXIT_CODE" value="true"/>
     <stringAttribute key="org.eclipse.jdt.launching.VM_ARGUMENTS" value="-Xmx512m &quot;-Dorg.freeplane.basedirectory=${workspace_loc:freeplane_debughelper}/../BIN&quot; &quot;-Djava.security.policy=${workspace_loc:freeplane_debughelper/security/freeplane.policy}&quot; -Dorg.freeplane.main.application.FreeplaneSecurityManager.disable=false -Dorg.freeplane.main.application.FreeplaneStarter.headless=false -Dorg.osgi.framework.bundle.parent=app -Dorg.freeplane.nosplash=false &quot;-Dorg.freeplane.builtin.scripts.dir=${workspace_loc:freeplane_plugin_script/scripts}&quot; &quot;-Dorg.freeplane.globalresourcedir=${workspace_loc:freeplane}/src/external/resources&quot; -Dsun.awt.disablegrab=true"/>
     <stringAttribute key="org.eclipse.jdt.launching.WORKING_DIRECTORY" value="${workspace_loc:freeplane_debughelper}/../BIN"/>
     <stringAttribute key="yk-options" value="&#13;&#10;additional-options2=onexit\=snapshot&#13;&#10;startup=0&#13;&#10;"/>
```
so with eclipse as IDE, freeplane can do restart like eclipse itself do.

dpolivaev:
I am sorry I don't think it can work because under windows we usually don't use the batch file and under mac we usually don't use the bash script. In both cases we use OS specific launchers generated by launcher frameworks.

If you want to make user experience better you could specify at each preference in the preferences.xml if it requires restart or not and so you can decide if the restart dialog is shown. You could add attribute `restartRequired=true`Actually in the most cases the option is applied immediately and restart is not needed. Therefore false could be the default (no extra attribute), and only in some rare cases it should be explicitly set to true.

danix800:
Are there any info about the OS specific launchers or packaging techniques for these two platforms? I still think that an action for restart is nice.

dpolivaev:
It is a long story.
We can schedule a meeting at the weekend if you like.

dpolivaev:
Personally, I see no strong use case for the restart action.
And if you do it, ideally you need to take care of all unsaved and untitled maps and also to make sure that you restore exactly the same windows independently on the preferences. Otherwise, restart does not make sense and you could get overwhelmed by user posts.

danix800:
Restart is just a plain old QUIT action with a different non-zero exit code signaling an auto relaunch request to external launchers. As long as there's no other exit code sensitive application of FP then there won't be other influence I think.

The thing I'm worried about is whether I can cover all the launching related cases by myself. For example when developing in eclipse whether the JDT request  will be accepted is undefined. For other launching techniques as long as we're in control then there won't be a problem from my perspective.

PR #216, opened 2021-06-07
Bugfix/overview interactions

danix800:
1. the first commit is for a confirmed bug (fix the bounds location when overview is attached to NW, otherwise it cannot be moved)
2.  the second commit is for overview size resetting (the original design is when attaching to the origin, second double click reset its size), but the fix is in doubt. Please give feedback if there's any problem.
3. the third commit is an improvement which can handle the case when overview is not attached to the very origin (0, 0) in its corresponding corner, which means in the original implementation, if the overview is moved away from the corner, the uncovering will not work. The hidden area is extended to the smaller corner relative to the underlying mapview so the original algorithm will think that the whole extended area is hidden, which makes the original algorithm works again.

Thanks for the uncovering implementation @dpolivaev .

### DIFF
--- a/freeplane/src/main/java/org/freeplane/view/swing/map/MapViewScrollPane.java
+++ b/freeplane/src/main/java/org/freeplane/view/swing/map/MapViewScrollPane.java
@@ -85,14 +85,10 @@ public class MapViewScrollPane extends JScrollPane implements IFreeplaneProperty
         public void scrollRectToVisible(Rectangle newContentRectangle) {
         	final Rectangle hiddenArea = hiddenAreaSupplier.getHiddenArea();
         	if(hiddenArea.width != 0 && hiddenArea.height != 0) {
-        		Point viewportLocation = new Point(0, 0);
-        		UITools.convertPointToAncestor(this, viewportLocation, JScrollPane.class);
-        		hiddenArea.x -= viewportLocation.x;
-        		hiddenArea.y -= viewportLocation.y;
         		final boolean isHiddenAreaAtTheLeft = hiddenArea.x == 0;
 				final boolean isHiddenAreaAtTheTop = hiddenArea.y == 0;
-				final boolean isHiddenAreaAtTheRight = hiddenArea.x + hiddenArea.width == getWidth();
-				final boolean isHiddenAreaAtTheBottom = hiddenArea.y + hiddenArea.height == getHeight();
+				final boolean isHiddenAreaAtTheRight = hiddenArea.x + hiddenArea.width == getParent().getWidth();
+				final boolean isHiddenAreaAtTheBottom = hiddenArea.y + hiddenArea.height == getParent().getHeight();
 				if(isHiddenAreaAtTheLeft || isHiddenAreaAtTheRight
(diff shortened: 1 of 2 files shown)


dpolivaev:
What do you mean by "uncovering"?
Probably you mean considering the map overview pane when the map is automatically scrolled. Have I got it?
As usual, I am starting my paid work now and I am getting back to Freeplane tasks probably on Friday.

danix800:
By "uncovering" I mean https://github.com/freeplane/freeplane/commit/5ee463aa6d4b09c16cc5a3ee7024ef031138bd2b `Adjust scrolling to move selected node outside of the map overview`.

freeplane/src/main/java/org/freeplane/view/swing/map/MapViewScrollPane.java
```diff
@@ -85,14 +85,10 @@ public void setBackgroundComponent(JComponent backgroundComponent) {
...
         	if(hiddenArea.width != 0 && hiddenArea.height != 0) {
-        		Point viewportLocation = new Point(0, 0);
-        		UITools.convertPointToAncestor(this, viewportLocation, JScrollPane.class);
-        		hiddenArea.x -= viewportLocation.x;
-        		hiddenArea.y -= viewportLocation.y;
         		final boolean isHiddenAreaAtTheLeft = hiddenArea.x == 0;
 				final boolean isHiddenAreaAtTheTop = hiddenArea.y == 0;
-				final boolean isHiddenAreaAtTheRight = hiddenArea.x + hiddenArea.width == getWidth();
-				final boolean isHiddenAreaAtTheBottom = hiddenArea.y + hiddenArea.height == getHeight();
+				final boolean isHiddenAreaAtTheRight = hiddenArea.x + hiddenArea.width == getParent().getWidth();
```
dpolivaev: I do not understand why this change was needed. Could you explain?
I understand my solution and I do not understand your change.

In my solution hidden area was intentionally given in JScrollPane coordinates and not in JViewport coordinates because MapOverview places JScrollPane and we do not know how JViewPort is internally placed by JScrollPane layout

danix800: Please take a look at the latest commit.

dpolivaev:
I am sorry I missed the issue and currently I do not know what it should solve and why it is important.
Can you help me or can I close it?

danix800:
The issues this PR tries to solve still exist, so I'll try to explain more precisely:

1. When overview is attached to top-left, it cannot be moved;
1. ... and cannot be resized either;
1. When the overview is attached to a corner (0, 0), nodes covered by overview can be displayed when selected, but when not attached to exactly (0, 0), these nodes won't show automatically.

All the above phenomena are observed on Debian 11/Gnome 3.38/JDK 1.8.0.281.

Hope this helps.

dpolivaev:
I am sorry, I do not understand what is currently going wrong.
Please explain which steps I should do to see the problem, 
which observations I can do, what is the difference between correct and observed behavior.
Add mind map test files, screenshots, or screen recordings if needed.

dpolivaev:
Closed because of no response from the committer

PR #218, opened 2021-06-07
drawing overview resizable border indicators on hovering

danix800:
This PR is for [this](https://sourceforge.net/p/freeplane/discussion/758437/thread/494dfaa629/#2f31/83ba/f68f/15fd/2f9f). It's purely for exercising swing/awt programming skills so whether or not it will be accepted are both OK with me.


dpolivaev:
I find the new version more animated and more distractive, so I share your doubts

PR #243, opened 2021-12-09
#2398 To copy an image to clipboard 

jakepu:


dpolivaev:
Thanks for your submission.
How does it handle svg images? I think they are not supported by ImageIO.

PR #248, opened 2022-01-27
Add support for reproducible builds

inglor:
Hi,

I'm the package maintainer for `freeplane` in Arch Linux and trying to make this package reproducible [1]. This will make the archives (JAR, ZIP etc) to omit timestamp information and maintain order within the archives between builds.

It also helps on gradle incremental builds to use more efficient the gradle internal cache.

[1]: https://reproducible-builds.org/


dpolivaev:
Merged, let's try it

PR #281, opened 2022-03-27
Corrected compilation warnings: .toURI() -> .toURI().toURL()

BeniaminK:
This change fixes the deprecation of File.toURL()
File.toURL is deprecated:
https://docs.oracle.com/javase/7/docs/api/java/io/File.html#toURL()

and few `show()` -> `setVisible(true)`


dpolivaev:
I think Freeplane intentionally uses the old method because it creates a different result.

Issue #522, opened 2022-08-02
extend API: add MindMap getListeners


macmarrum:
Scripting API provides methods on MindMap to addListener and removeListener.
Adding a listener is straight-forward. Yet, to remove a previously added listener, User needs to provide its reference. And there is little possibility to keep the reference in the script that adds a listener.

This is to request an extension of the API so that MindMap exposes a list of listeners, so that User can finds its listener and supply it to removeListener.

PR #523, opened 2022-08-02
extend API: add MindMap getListeners

macmarrum:
implement #522


dpolivaev:
Please check my latest commit to your branch and let me know if it is all right so that I can merge it into the freeplane repo.

macmarrum:
Thanks, good changes
FP compiled and the new functionality executed OK
You can go ahead with the merging

PR #631, opened 2022-09-04
Update README.md with ubuntu/debian instructions

edzob:
Added ubuntu/ debian install instructions since sourceforge is not applicable for these platforms.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,3 +47,37 @@ Every contributor or team member freely decides what task they are going to work
 Where to Download
 =====================================
 Download and install the latest version over at [Sourceforge](https://sourceforge.net/projects/freeplane/files/).
+
+How to Install in Ubuntu/ Debian
+=====================================
+Freeplan is part of default repositories in 
+[Ubuntu](https://launchpad.net/ubuntu/+source/freeplane) and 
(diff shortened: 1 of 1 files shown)


fnatter:
hu edzob,

note that there is a .deb for Debian and Ubuntu (works for both and almost all versions
because all dependencies are included)
on [sourceforge](https://sourceforge.net/projects/freeplane/files/freeplane%20stable/): 
freeplane_VERSION~upstream-1_all.deb. Do you want to mention this?

It is always up-to-date unlike the official Debian/Ubuntu package(s). Don't get me wrong: I prefer
the official "main" archive(!), but the lack of recent gradle (and some other hurdles) made it
impossible to maintain the official package(s). If someone wants to help, please step in!
I'll be happy to provide mentoring and help!

Cheers and Best Regards,
Felix

edzob:
Adding is no problem, but i am not that experienced with PR, so I need to do some research if I need to cancel or can add a commit to a PR.

dpolivaev:
I think that we should not recommend installation from official Debian or snap repositories because they are outdated. Therefore I am closing this pull request now.

Generally, if you push a new commit to the branch having a pull request, it is automatically added to the pull request because it always considers the whole source branch.

Issue #649, opened 2022-09-10
Freeplane won't start up saying: "This application requires a Java Runtime Environment 11 - 17"


PontiacCZ:
**Describe the bug**
I have just installed 64bit Freeplane 1.10.3 using **Freeplane-Setup-1.10.3.exe**. When I launch the application I get a window saying:

_This application requires a Java Runtime Environment 11 - 17_

![image](https://user-images.githubusercontent.com/9153595/189484600-e8c48c99-d86e-45e0-a212-c20fa8bfa759.png)

After I click the OK button a Java webpage opens offering download 64bit Java Version 8 Update 341.

![image](https://user-images.githubusercontent.com/9153595/189484748-24c2583b-944c-473f-ac6f-5b386771b05b.png)

BUT I already have Java 8 update 341 thanks to automatic updates:

![image](https://user-images.githubusercontent.com/9153595/189484882-5b8a0ed5-ea1a-4ccf-86a6-1e50e042131f.png)

![image](https://user-images.githubusercontent.com/9153595/189484989-278fb7c1-adc3-4c74-a4b2-2349fccdc66f.png)

And what's more, before uninstalling previous Freeplane version 1.9.12 (which had come from _Freeplane-Setup-1.9.12.exe_) I was using it. It ran smoothly and opened my recent mindmap just fine.

**Desktop:**
 - OS: Windows 10 Pro
 - Version 21H2
 - Java Version 8 update 341

lukailicgithub:
I have the exact same issue.

Can someone help us here?

Merci!

euu2021:
One easy solution is using the version that comes with Java: Freeplane-Setup-with-Java-1.10.4.exe (edit for clarification: or any newer version that comes with Java, including the preview versions)

![image](https://user-images.githubusercontent.com/77707706/189505713-b5752400-e881-41cc-ab53-dae9a62300d8.png)

For more solutions, see:
#275 
#365 
#462

euu2021:
Clarification: in my example above, I used the version 1.10.4 simply because it was the latest stable version by the time that I was writing. The important thing here is that you use the version that already comes with Java. The same effect can be achieved by using the latest preview version, for example.

See more in [this comment](https://github.com/freeplane/freeplane/discussions/652#discussioncomment-3618720)

PontiacCZ:
I have just installed fresh new version 1.10.4, but the issue stays the same.

I know about Freeplane with bundled Java. This report is about the installer without Java.

[...]

PR #655, opened 2022-09-11
change download url for java

macmarrum:
When Launch4j redirects to http://java.com/download, which offers only JRE 8, User is lost and goes on to download and install JRE 8, despite the initial warning: "The application requires a Java Runtime Environment 11 - 17"

The new link redirects to Zulu JRE 15, like the one used in Freeplane-Setup-with-Java.exe


macmarrum:
Related to #649

dpolivaev:
I tried to find it but failed

PR #660, opened 2022-09-12
Update README.md

euu2021:
I added a comment about the documentation at the beginning of the page.


euu2021:
Some people may not know what "documentation" is (mainly, when using autotranslation), so I also added a description.

PR #726, opened 2022-10-05
Feature 700/insert condi

macmarrum:
1. fix a bug in undo add/insert
2. add/insert/remove conditionalStyle were apparently created to be used by the Manage Conditional Styles Action (dialog) and lack the call to nodeChanged, because it is done after the dialog is closed.
This PR adds methods which call nodeChanged. They will be used by the upcoming scripting API.


dpolivaev:
Hello, I have just seen this PR.
Is it still relevant or should I close it?

macmarrum:
Not needed any more

Issue #896, opened 2022-12-19
using &nbsp; within freeplane's XML files causes parsing error


nnako:
**Describe the bug**
Any Freeplane file is expected to be a "valid" XML file, following the rules defined by that format standard. The file `freeplaneFunctions.mm` which can be opened using the `help` menu uses both, `&nbsp;` as well as the proper `&#160;` for encoding non-breaking-spaces. Although the respective helper mindmap is displayed properly within the Freeplane editor, the parsing using a standard XML parser fails (at least in case of using the parser from the widespread Python package `lxml`). Using these non-defined `&nbsp;` elements within XML files renders them "invalid" although they are still "well-formed". Maybe, the invalid content might have been pasted from a pure XHTML source without having XML in mind?

please see a [corresponding bug report on lxml project's issue tracker](https://bugs.launchpad.net/lxml/+bug/1936579).

**To Reproduce**
1. open Freeplane
2. open Help -> Tutorial
3. open Tools -> Open current map directory
4. copy the path to the file freeplaneFunctions.mm
5. pick the file using the command line and try to parse it using Python's `lxml` package

**Expected behavior**
no errors on the command line

**Mind map files**
[freeplaneFunctions.zip](https://github.com/freeplane/freeplane/files/10262655/freeplaneFunctions.zip)

**Screenshots**
```bash
Traceback (most recent call last):
...
  File "src/lxml/etree.pyx", line 3538, in lxml.etree.parse
  File "src/lxml/parser.pxi", line 1876, in lxml.etree._parseDocument
  File "src/lxml/parser.pxi", line 1902, in lxml.etree._parseDocumentFromURL
  File "src/lxml/parser.pxi", line 1805, in lxml.etree._parseDocFromFile
  File "src/lxml/parser.pxi", line 1177, in lxml.etree._BaseParser._parseDocFromFile
  File "src/lxml/parser.pxi", line 615, in lxml.etree._ParserContext._handleParseResultDoc
  File "src/lxml/parser.pxi", line 725, in lxml.etree._handleParseResult
  File "src/lxml/parser.pxi", line 654, in lxml.etree._raiseParseError
  File "/freeplaneFunctions.mm", line 135
lxml.etree.XMLSyntaxError: Entity 'nbsp' not defined, line 135, column 101
```

**Additional context**
nothing relevant, as outside of Freeplane itself.

dpolivaev:
Preview 1.11.1_06

PR #897, opened 2022-12-19
Issue #896

nnako:
replaced all occurrences of `&nbsp;` with `&#160;` within the `freeplaneFunctions.mm` file renders this file as proper XML while not changing any view or behavior when loaded and viewed within the Freeplane editor.


dpolivaev:
I fixed it, thank you

PR #1043, opened 2023-03-08
Support gradle 8.x

inglor:
Probably backport in 1.11.x as well since both branches are affected.

version 10.0.0 changed the namespace
version 11.0.0 adds supports for gradle 8.x

https://github.com/nebula-plugins/gradle-ospackage-plugin/releases

@dpolivaev Not sure if you want to handle the `dependsOn` that way, that's why it's a separate commit.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -10,7 +10,7 @@ buildscript {
 }
 
 plugins {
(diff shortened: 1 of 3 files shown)


inglor:
Closed in favour of #1124

PR #1243, opened 2023-06-18
decode at for execution and uri for editing

macmarrum:


dpolivaev:
I am not sure when the strings are encoded and when they are not. Actually, in Java URI is always encoded but for URL it depends. Therefore I am not sure if the fix is needed and if it never makes any trouble.

macmarrum:
How can I help to increase your level of confidence?

dpolivaev:
Hello, I am sorry for responding to you when I actually was having no time myself.

Could you please explain the reasons behind each change you made? 

Seeing only the changes and without understanding the cases behind it I can not see which problem or problems are solved and if the solution is correct. If it is a bug fix please describe the bugs.

For me, it is also not clear why the complete text is always URL-encoded, that you should decode it before showing it to the user, and when it is encoded again.

macmarrum:
I have analyzed the cases some more and realized I've made some assumptions, therefore I am withdrawing this PR. I will create a new one.

PR #1313, opened 2023-07-21
Feature/markdown hard line breaks

macmarrum:
- fix htmlToPlain to retain trailing NBSPs so that trailing spaces (nbsp) are respected, which is one of the methods to indicate a hard line break in markdown
- bump markedj version because 1.0.18 includes a fix for hard line breaks by backslash at EOL -> https://github.com/gitbucket/markedj/pull/41


dpolivaev:
Preview 1.11.5_19

PR #1315, opened 2023-07-23
fix NPE when text is null

macmarrum:
Exception in thread "AWT-EventQueue-0" java.lang.NullPointerException: Cannot invoke "String.equals(Object)" because "threadLocalCachedHtmlToPlain.input" is null
	at org.freeplane.core.util.HtmlUtils.htmlToPlain(HtmlUtils.java:116)

### Test Case
[api-groovy-tutorial.xi.zip](https://github.com/freeplane/freeplane/files/12139611/api-groovy-tutorial.xi.zip)
- Enable Map Conditional Style in the attached mind map


dpolivaev:
DO you think I should create a new preview version or can it wait?

macmarrum:
I've been running Freeplane with this code and so far it works fine.

I believe there is little chance of side effects, therefore I think the preview for a wider audience can wait.

Unless you've had some other reason for asking about a preview version?

PR #1324, opened 2023-08-04
fix: make sure #at(...) is decoded

macmarrum:
this fixes the issue that special characters like spaces make the at-path invalid

[test-FP1243-decode at for execution and uri for editing.mm.zip](https://github.com/freeplane/freeplane/files/12263132/test-FP1243-decode.at.for.execution.and.uri.for.editing.mm.zip)


dpolivaev:
I don't understand where special characters come from and where they are encoded. 
Can you please give me an example?
Do you want the user to encode them or should Freeplane encode them when some URL containing the reference is created?

dpolivaev:
Could you find some time to talk about this story?

macmarrum:
Sure thing.
I should be back in 1h and will try to catch you then on Discord.

dpolivaev:
It is getting too late. If you are not available right now we can meet on another day.

PR #1353, opened 2023-08-20
Rename tabs

dpolivaev:


freeplane/src/main/java/org/freeplane/main/application/MapViewDockingWindows.java
```diff
@@ -208,9 +209,37 @@ public void internalFrameClosing(InternalFrameEvent e) {
...
 
 	}
 
+	private void addTabsPopupMenu(DockingWindow dockingWindow){
+		dockingWindow.setPopupMenuFactory(new WindowPopupMenuFactory() {
+		   public JPopupMenu createPopupMenu(DockingWindow window) {
+			   if(window instanceof ConnectedToMenuView) {
+				   return null;
+			   }
+			   JPopupMenu menu = new JPopupMenu(window.getTitle());
```
dpolivaev: Have you added an import for JPopupMenu ?

freeplane/src/main/java/org/freeplane/main/application/MapViewDockingWindows.java
```diff
@@ -208,9 +209,37 @@ public void internalFrameClosing(InternalFrameEvent e) {
...
+			   JMenuItem menuItem = new JMenuItem("Rename tab (temporarily)");
```
dpolivaev: Add translation keys and English versions to /freeplane/src/viewer/resources/translations/Resources_en.properties and use them through TextUtils.getText(String, String). Don't include any texts directly in the code because of translations.

freeplane/src/main/java/org/freeplane/main/application/MapViewDockingWindows.java
EdoFro: I did, but IntelliJ changed all single imports for one `import javax.swing.*` 
And when I saw that I let that part out of the commit, missing to add the JPopMenu import.

I'm making the changes in my branch and will to do a pull request to learn the correct cooperation flow.

thanks again

dpolivaev: You do not need to open any new pull request because all changes on your branch are automatically added to this pull request

EdoFro: ok. understood.

EdoFro:
I think now I've included the whole changes. 😅

dpolivaev:
When I run your version and right-click on the area near the tabs, I can see a new pop-up menu that allows me to show the input for renaming the tab. However, if I enter any text there, neither the tab title nor the frame title is changed. Entering text in the input field has no effect.

I would also expect that the pop-up menu appears when I click on the tab and not in the free space to the right of the tabs.

renameTabInput

EdoFro:
> When I run your version and right-click on the area near the tabs, I can see a new pop-up menu that allows me to show the input for renaming the tab.

I didn't test that. I will change it to work only on the tabs.

Currently it works only on Tabs from **grouped windows**. That's why it doesn't show anything when right clicking on "one window" tabs:

image

dpolivaev:
If I right-click on the tabs itself, no popup is shown at all. I tried two different Looks and Feels.

EdoFro:
> If I right-click on the tabs itself, no popup is shown at all. I tried two different Looks and Feels.

Did you group tabs?
it only works in tabs that have multiple child windows.

dpolivaev:
No, I use only one Freeplane window for all maps.
Can you fix it also for this case?

EdoFro:
Ok. I have now some busy days, but after that I will try it again.

I'm not at my computer right now, so I will copy here a link to remember the methods I have to look at:
https://github.com/freeplane/freeplane/blob/6bf842791d968c9e2a35c4432627d37e2a1ac9d2/freeplane/src/main/java/org/freeplane/main/application/MapViewDockingWindows.java#L457-L479

Regards,

edo

EdoFro:
I commit a little change, but it is no ready yet. I have to test it and add other changes. It was just to write down the idea and not forget it.

Please wait before testing.

EdoFro:
> No, I use only one Freeplane window for all maps.
> Can you fix it also for this case?

NOTE TO MYSELF:
just for the record (and to look at that Freeplane code):

this script changes a mind map's tab's title:
```groovy
node.mindMap.setName("My custom Name")
```

It doesn't show the change in the title right after runing it, but if you change the focused mapview, then it will change.

Strange is that the getter returns the original name:
```groovy
node.mindMap.getName()
```

To return to the orignal title:
```groovy
node.mindMap.setName(node.mindMap.getName())
```
and this throws an error:
```groovy
node.mindMap.setName(null)
```

dpolivaev:
The Interface and logic of DockingWindows are often not intuitive. When I get time I can have a look, but it is not my highest priority at the moment.

Here you can see that your recent change broke the build. It does not compile now.

EdoFro:
> The Interface and logic of DockingWindows are often not intuitive. 

Completelly agree!!!!

>When I get time I can have a look, but it is not my highest priority at the moment.

Don't worry. If I need help or any advice I will explicitly ask for it. I also think this feature is not a priority. I ´m doing it to learn and not to mess with important ones.

> 
> Here you can see that your recent change broke the build. It does not compile now.

OK. I will look at it when I get back to my computer.

I will let you know when I think it is good enough for you to test again.

Thanks,

edo

EdoFro:
> No, I use only one Freeplane window for all maps.
> Can you fix it also for this case?

Done.

I think it is ready for testing. It works in my computer.

Regards,

edo

EdoFro:
Hi @dpolivaev,
I worked on this branch again.
I commited some changes, but this time I was very careful in adding meaningful titles to each of them.

Here are two videos
In the first one:
- renaming a tab
- pop up menu shows the previous name as predefined text
- if cancelled or input is "" then it returns to its original name
- renaming all the tabs
- grouping the windows
- renaming the tabs from the grouped window

https://github.com/freeplane/freeplane/assets/35700575/e46edddf-2384-4742-9473-1a0d33124450



in the second one:
I close Freeplane and then I reopen it and the tab names are retained

https://github.com/freeplane/freeplane/assets/35700575/45392578-a3ee-4a1b-ac31-6e48678f7084

EdoFro:
@dpolivaev 

Dear Dimitry,
I have been using this version of Freeplane for the last week and it has not presented any problems.
I think it could be added to a preview version so that other users could test it and give their opinion.

Please let me know what you think.

Best regards,

edo

dpolivaev:
At first I want to replace CustomWindowTitleProvider class by method reference
Could you please give me write access to your repository? It is needed because if you created the pull request I had it. But because I tried to simplify things to you I have no write access to the branch and so I can not show you what I mean. Creating extra pull request from me to you is too much complexity, I just want to have the write access.

And then I am going to tell you which parts of the code I do not understand.

dpolivaev:
Basically, I do not understand working with '*' characters in 

`````java
    public String getTitle(DockingWindow window){
        String windowName = (String) window.getClientProperty(MapViewDockingWindows.CUSTOMIZED_TAB_NAME_PROPERTY);
        if (windowName == null || (window instanceof ConnectedToMenuView)) {
            return SimpleDockingWindowTitleProvider.INSTANCE.getTitle(window);
        } else {
            boolean dirty = false;
            int childWindowsCount = window.getChildWindowCount();
            for (int i = 0; i < childWindowsCount; i++) {
                dirty = dirty || window.getChildWindow(i).getTitle().endsWith("*");
            }
            return windowName + (childWindowsCount>1?" (" + childWindowsCount + ")":"") +(dirty?" *":"");
        }
    }

`````

and in 

`````java
	private void loadInitialCustomTabNames(){
		for (Component mapViewComponent: mapViews) {
			if (mapViewComponent instanceof MapView ) {
				View containingDockedWindow = getContainingDockedWindow(mapViewComponent);
				if(containingDockedWindow != null) {
					String oldTitle = containingDockedWindow.getViewProperties().getTitle();
					MapView mapView = (MapView)mapViewComponent;
					if(oldTitle != null && !oldTitle.equals("") && !oldTitle.equals(mapView.getName())){
						if(oldTitle.endsWith(" *")) {
							oldTitle = oldTitle.substring(0,oldTitle.length()-2);
						}
						mapView.putClientProperty(CUSTOMIZED_TAB_NAME_PROPERTY, oldTitle);
					}
				}
			}
		}
		initialTabNameLoadingWasDone = true;
	}

`````

I do not understand the ideas behind it.

**Comment by dpolivaev** [comment]
I have seen that the tab names are persisted between Freeplane starts. If it is the case, why are they called "temporary"?

**Comment by dpolivaev** [comment]
Actually, if I rename the tab for a map that does not have other views, the tab name is kept after the restart. But if there are multiple views of the same map, only one of them is loaded after the restart, and then we do not know if it was the renamed one or another one.

**Comment by EdoFro** [comment]
> Could you please give me write access to your repository?

Ok. done (I think)

I did this, I don't know if I have to do some extra stuff.
image

**Comment by EdoFro** [comment]
> At first I want to replace CustomWindowTitleProvider class by method reference

I looked into how infonode's docking windows worked and they use a title provider to name the tabs that group windows together. So even if you change the name, when it call the standard simpleTitleProvider again, it changes the name to a concatenation of the names of its children. 
The only way to avoid this is to assign a custom title provider to the window.

I explain this better here (includes links to the source code):
https://github.com/freeplane/freeplane/issues/865#issuecomment-1648768449

**Comment by EdoFro** [comment]
> Basically, I do not understand working with '*' characters in
> 

First the idea:

1. if I have a tab that has childwindows (in this case 'CustomMenus' and 'Mo')
2. the text in the tab is a concatenation of the labels of the child windows ('CustomMenus, Mo):
image

3. if I rename this group tab, (for example to 'Group'), it is not easy to recognize that is a group tab, that's why I decided to add the quantity of child windows at the end ( in this case it changes to 'Group (2)' )
image

4. if one of the childwindows gets "dirty", that was noticeable before when having the concatenations of tab labels as the group tab label.
image


5. So I decided that if any of it's child windows has an asterisk at the end then it should show an asterisk at the end itself to show that are dirty windows in its descendants.
image




That's what is in this code:

```java
    public String getTitle(DockingWindow window){
        String windowName = (String) window.getClientProperty(MapViewDockingWindows.CUSTOMIZED_TAB_NAME_PROPERTY);
        if (windowName == null || (window instanceof ConnectedToMenuView)) {
            return SimpleDockingWindowTitleProvider.INSTANCE.getTitle(window);
        } else {

            // is any of it's child windows "dirty"? (has an '*' at the end?)

            boolean dirty = false;
            int childWindowsCount = window.getChildWindowCount();
            for (int i = 0; i < childWindowsCount; i++) {
                dirty = dirty || window.getChildWindow(i).getTitle().endsWith("*");
            }

           //if yes, add an " *" at the end of the label

            return windowName + (childWindowsCount>1?" (" + childWindowsCount + ")":"") +(dirty?" *":"");
        }
    }
```

EdoFro:
> 
> and in
> 
> ```java
> 	private void loadInitialCustomTabNames(){
> 		for (Component mapViewComponent: mapViews) {
> 			if (mapViewComponent instanceof MapView ) {
> 				View containingDockedWindow = getContainingDockedWindow(mapViewComponent);
> 				if(containingDockedWindow != null) {
> 					String oldTitle = containingDockedWindow.getViewProperties().getTitle();
> 					MapView mapView = (MapView)mapViewComponent;
> 					if(oldTitle != null && !oldTitle.equals("") && !oldTitle.equals(mapView.getName())){
> 						if(oldTitle.endsWith(" *")) {
> 							oldTitle = oldTitle.substring(0,oldTitle.length()-2);
> 						}
> 						mapView.putClientProperty(CUSTOMIZED_TAB_NAME_PROPERTY, oldTitle);
> 					}
> 				}
> 			}
> 		}
> 		initialTabNameLoadingWasDone = true;
> 	}
> ```
> 
> I do not understand the ideas behind it.

I noticed that in the case that:
- there where unsaved mindmaps
- and you close Freeplane

It appears the dialog to decide "save, ignore, cancel":
image

you can save or not, but the information of the windows saved to auto.properties includes the "\*" (in the image 'Mo \*'), so next time you open Freeplane, it takes that name with the "\*" as the new tab label.
and when the mindMap gets dirty it has two "\*" ( in this case "Mo \* \*" )

That is why of this lines:

>```java
> 						if(oldTitle.endsWith(" *")) {
> 							oldTitle = oldTitle.substring(0,oldTitle.length()-2);
> 						}
>```

 *If the oldtitle (the one loaded from the auto.properties file) has already an '\*' at the end, then drop the last two letters.*

EdoFro:
> I have seen that the tab names are persisted between Freeplane starts. If it is the case, why are they called "temporary"?

The tab names from single windows persist, but the ones from grouped windows do not:
- they get reseted when changing the group's layout
- they do not persist between Freeplane starts.

Maybe that "temporary" message should only be shown to the grouped windows tabs:
- pop up shows one message when it's a tab from a single window and another one when it's a tab from a grouped window

What do you think?

EdoFro:
> Actually, if I rename the tab for a map that does not have other views, the tab name is kept after the restart. But if there are multiple views of the same map, only one of them is loaded after the restart, and then we do not know if it was the renamed one or another one.

I haven't test this one. I don't know if it can be solved. That's all in the infonode logic. I just did some tricks to try to get the best outcome of that.

Today I'm going to be out of town. I will have my phone, but no laptop.

I tried to answer everything the most clearly possible. I hope it was ok.

Best regards,

edo

dpolivaev:
I do not even know how to create a group tab. Could you please explain it?
Anyway, I removed the word "temporary" because it is not always temporary. 
I don't want to block the process and I am merging it now.

EdoFro:
> I do not even know how to create a group tab. Could you please explain it?

🤣 I assumed all along that you knew about this. Sorry, I didn't imagine it was something new to you, but now I understand it, because it is a functionality of a library external to Freeplane. 

1. normal case:
if you select a tab and drag'n drop it, then you can change the order of the tabs or the layout of the windows on the screen, but no group tab is created.

2. but, ( 3 or more views open) 
if, instead, you drag'n drop another tab (not the selected one) and drop it into the selected view, then a new group is created that includes the selected view and the view you drop into it.

I will add an example video i a next post.

EdoFro:
In the video:
1. I select view "CustomMenu" (click on its tab)
2. I drag'n drop tab "Mochila" into view "CustomMenu"
3. Now they are both part of a new group view
You can change its layout by moving the tabs inside the view


https://github.com/freeplane/freeplane/assets/35700575/beec42f3-44cf-4a7f-aeee-db044882b3b7

dpolivaev:
Got it. In the following configuration, it shows 2 but it should show 3
group3

EdoFro:
> In the following configuration, it shows 2 but it should show 3

yes. it counts the direct child windows of the group tab, not all descendant views. In this case there are two windows: top an bottom, and the bottom one is splitted in two (left and right).

I think it would be better if it counts all descendant views, but I didn't want to make it more complex before solving the other issues.

I will try to do it in a next commit.

EdoFro:
I added some code that solves this issue:
https://github.com/freeplane/freeplane/pull/1471

euu2021:
I'm testing the new renaming tabs feature for the first time, now. It will be useful. Thanks!

aaa1386:
Thank.

I was happy with the new changes in the Freeplane.

An idea to complete what happened: When a node in the new map view is used to focus on a branch by the "jump in" option, its name should be the text tab of the first node.

EdoFro:
> Thank.
> 
> I was happy with the new changes in the Freeplane.
> 
> An idea to complete what happened: When a node in the new map view is used to focus on a branch by the "jump in" option, its name should be the text tab of the first node.

as a new feature request in #1484

PR #1382, opened 2023-09-05
image uri and file fixes

macmarrum:
- fix for `node.externalObject.uri = new URI('relative.png')`
- fix for `node.link.file` when `node.link.uri == new URI('/no-schema/absolute/path/file.txt')`


dpolivaev:
What do you think about using `path.toAbsolutePath().normalize()` as suggested in https://stackoverflow.com/questions/45938856/how-can-i-determine-the-canonical-path-of-a-file-without-following-symbolic-link ?

macmarrum:
I have read up on the subject and I learned that `normalize()` can have undesired effects when the pre-`..` segment is a symbolic link.

https://docs.oracle.com/javase/8/docs/api/java/nio/file/Path.html#normalize--
> Eliminating ".." and a preceding name from a path may result in the path that locates a different file than the original path. This can arise when the preceding name is a symbolic link.

Therefore I'd like to leave the path as is, without normalization.

PR #1414, opened 2023-09-16
fix image uri when linkType==relative

macmarrum:
After running more tests, I found a bug. It caused Pack And Go-Go to stop working when OptionPanel.links == "relative". I realized that a different approach was needed to make it work, therefore my original changes for pasteImage, from PR #1382, are no longer required and can be reverted.

I think that for scripting purposes, it's best to allow User to decide what URI they want to set for externalObject – be it absolute or relative, therefore I extracted a method to be used for this purpose.
Also, this is a requirement for Pack And Go-Go to work correctly.

### DIFF
--- a/freeplane/src/main/java/org/freeplane/view/swing/features/filepreview/ViewerController.java
+++ b/freeplane/src/main/java/org/freeplane/view/swing/features/filepreview/ViewerController.java
@@ -778,27 +778,20 @@ public class ViewerController extends PersistentNodeHook implements INodeViewLif
 			return false;
 		}
 		File file = new File(uri.getPath());
-		final File mapFile = targetNode.getMap().getFile();
-		if (!file.isAbsolute()) {
-			if (mapFile == null) {
-				showWarningMapNotSaved();
-				return false;
-			}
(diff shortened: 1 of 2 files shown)


dpolivaev:
@macmarrum 
I have merged your pull request.
I do not understand what it does and do not want to spend time on it.
I have no doubts about your skills, responsibility, and goodwill.
So I suggest that you merge any changes by yourself instead of creating pull requests.
If I get any remarks or ideas related to implementation or design and if any bugs are discovered, I believe that you fix issues as soon as possible for you.
Only if the bugs or issues are serious and you have no time to rework it I would revert your changes.
To simplify possible reverts, let us continue using feature branches and merging everything without rebase so that the changes can technically be reverted quickly if needed.

dpolivaev:
@macmarrum  I have added you to the team having the repo write access. Maybe you need to accept the invitation.

macmarrum:
@dpolivaev
Thanks for adding me to the team :+1:
Going forward, I'll merge changes without a PR, from a feature branch, as suggested.

PR #1507, opened 2023-10-31
adds mindMap.getStyleNodes method to API

EdoFro:
This PR adds a method to the mindMap API to get access to the styleNodes in the styleMap.

script example :
```groovy
import org.freeplane.api.StyleGroup

def map = node.mindMap

println "Predefined styles       : ${ map.getStyleNodes( StyleGroup.PREDEFINED       ) } \n"
println "User defined styles     : ${ map.getStyleNodes( StyleGroup.USER_DEFINED     ) } \n"
println "Automatic layout styles : ${ map.getStyleNodes( StyleGroup.AUTOMATIC_LAYOUT ) } \n"

```

outputs :
```
Predefined styles       : [(Node) Default, (Node) Details, (Node) Attributes, (Node) Note, (Node) Floating node, (Node) Selected node] 

User defined styles     : [(Node) Important] 

Automatic layout styles : [(Node) Root, (Node) Level 1, (Node) Level 2, (Node) Level 3, (Node) Level 4, (Node) Level 5, (Node) Level 6, (Node) Level 7, (Node) Level 8, (Node) Level 9, (Node) Level 10] 


```

### DIFF
--- a/freeplane_api/src/main/java/org/freeplane/api/MindMapRO.java
+++ b/freeplane_api/src/main/java/org/freeplane/api/MindMapRO.java
@@ -2,6 +2,7 @@ package org.freeplane.api;
 
 import java.awt.Color;
 import java.io.File;
+import java.util.List;
 
 /** The map a node belongs to: <code>node.map</code> - read-only. 
  * 
@@ -38,4 +39,13 @@ public interface MindMapRO {
 
 	/** @since 1.11.1 */
 	ConditionalStyles getConditionalStyles();
+
+	/** returns a list with the nodes of the specified group from the styleMap
+	 *
+	 * @param group StyleGroup.PREDEFINED, USER_DEFINED or AUTOMATIC_LAYOUT
+	 * @since 1.11.8
+	 * @return list of nodes
+	 */
+	List<? extends Node> getStyleNodes(StyleGroup group);
+
 }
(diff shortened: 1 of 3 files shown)


dpolivaev:
It is tricky, and it is hard to get it right because if you return style nodes in the script API, they and the structure of the style map can be modified in any unpredictable way. But Freeplane expects the style map to have a certain structure. Also, even if style nodes are updated correctly, Freeplane does not send an event telling that styles have changed, because the script treats them as usual nodes.

dpolivaev:
Therefore in order to do it correctly you probably need new interfaces and "proxies" specific to style nodes that do not extend `org.freeplane.api.Node`

dpolivaev:
If you decide to work on it, please submit a new pull request.

EdoFro:
> Therefore in order to do it correctly you probably need new interfaces and "proxies" specific to style nodes that do not extend `org.freeplane.api.Node`

OK. Agree. 👍🏼 

If I understand correctly, I have to create a new api (let's say "StyleNode") that is similar to api.Node, but has only the methods I need (getting and setting formats like Background, Edge or Border)

Can the "StyleNodeProxy" then call the NodeProxy methods that it needs or do I have to replicate them in the new proxy?

thanks,

edo

EdoFro:
> If you decide to work on it, please submit a new pull request.

OK

I have some use cases for this.
In one of them, I only need the list of names (Strings) of the User Defined styles. I will submit that first and leave the more complex ones for later.

Best regards,

edo

dpolivaev:
> Can the "StyleNodeProxy" then call the NodeProxy methods that it needs or do I have to replicate them in the new proxy?

Yes, one proxy can wrap another one. But in any case, extra logic should be implemented because it depends on the style which properties may be changed and which may not.

PR #1536, opened 2023-11-17
Adding node with link to menu item

EdoFro:
This PR adds a new feature to Freeplane:
- using a  Hot Key combination to insert a node linked to the command under the mouse pointer

 It uses the same logic as the "Text change" feature for Weblate

It adds a new menu command to set a new Hot Key combination.

image

## How to:

1. Select a node
2. Set the mouse over a menu item or a toolbar button
3. Use the Hot Key combination
4. a new node is inserted as a child from the selected node
    - it has an hyperlink to the menu item
    - it takes the menu item label as its text

## example video

https://github.com/freeplane/freeplane/assets/35700575/c290e974-fbc5-46f2-a40e-22a99b58d8e1


EdoFro:
I took some decisions and I want to know if they are ok:

1. I created a new package for this feature: `package org.freeplane.core.ui.commandtonode;`
I created it into **org.freeplane.core.ui**, because the `textchanger` package was there and they are very similar. 

2. I didn't know where was the correct place to add the `modeController.addAction(new ...Action)` code, and decided to add it to the **MLinkController** `createActions()` method.
3. I put the new feature in the `insert->link` menu, because it's related with the "add hyperlink to menu item" feature.
4. I decided to use `Ctrl + Shift + F2` as Hot Key, because it can be typed with one hand (left hand), because most people uses the mouse with the right hand.

Any opinions?

Bye,
edo

dpolivaev:
I understand your points. However, I want to raise some concerns and discuss an alternative solution.

Your current solution will not work on MacOS because on MacOS the menu is implemented by the OS and SwingUtilities will not find it. I would actually add "Insert node linked to menu command" as the new action and implement is similar to SetAcceleratorOnNextClickAction and implementing extending the logic in AccelerateableAction.actionPerformed(ActionEvent).

It would make the menu look better (I like a menu item name "Insert node linked to menu command" more than "Set 'Insert node linked to menu command' hot key") and allow it to work on Macs too.

EdoFro:
> Your current solution will not work on MacOS because on MacOS the menu is implemented by the OS and SwingUtilities will not find it

Oh! I didn't knew that.
I based my code in the logic from the Weblate "Change text" package. It may have this problem too. It's not really a feature from Freeplane for the mindMaps, it's an aid for people who want to collaborate with the translation. So, I think, it's not so important if it works in all OS.

>  I would actually add "Insert node linked to menu command" as the new action and implement is similar to SetAcceleratorOnNextClickAction and implementing extending the logic in AccelerateableAction.actionPerformed(ActionEvent).

Ok. I will look at it and try to do it that way.

> It would make the menu look better (I like a menu item name "Insert node linked to menu command" more than "Set 'Insert node linked to menu command' hot key") and allow it to work on Macs too.

I totally agree. I didn't like that text neither, but I tried to stay as simmilar as possible to the "Weblate text change" package.
I thought it would be easier for you to review it.

Thanks,
edo

EdoFro:
I think it would be correct to close this PR, because the new version may be quite different and it may take me some time to have it ready.

PR #1539, opened 2023-11-18
Copy attributes when copying styles between maps

EdoFro:
I think now it is ok:
- it removes the old attributes from the target node (if any)
- it copies the attributes from the sourceNode to the targetNode

I did multiple tests:
- multiple combinations of source and target nodes:
   - with / without icons
   - with / without attributes

and it works.

BR

edo


dpolivaev:
MapStyleModel.copyStyle is called from undoable operation, but mAttributeController.copyAttributesToNode and mAttributeController.performRemoveRow are undoable on their own. Therefore I need to check if your solution works fine or if it needs another solution. I am sorry that I do not have time this weekend to check it.

EdoFro:
Dear Dimitry,
It's no problem at all.
I had some time now so I continued with this task, but I don't expect that you can take a look at it immediatelly ( or even this weekend).
Please do not worry.

Thanks for your time 😉 👍🏼

dpolivaev:
Could you please test if my at your branch changes work fine?

EdoFro:
> Could you please test if my at your branch changes work fine?

It works perfect.

Thanks

EdoFro:
@dpolivaev,
I was thinking that adding formulas to attributes in template maps could be a way to add malicious code to many maps and users.
That's why I added some code to neutralize any formula values in attributes copied as part of a style node.

I will send a new PR.

edo

PR #1544, opened 2023-11-22
Update scripting api generator

EdoFro:
Hi Dimitry,
just some changes to the apiGenerator.groovy file:

- Added missing Classes (that were present in the API html documentation, but not here)
- Added enums and constants from classes
- Classes sorted alfabetically
- other little changes and bug corrections

BR

edo

### DIFF
--- a/freeplane_plugin_script/scripts/apiGenerator.groovy
+++ b/freeplane_plugin_script/scripts/apiGenerator.groovy
@@ -262,6 +262,17 @@ def createChild(Proxy.Node parent, text, link) {
     return result
 }
 
+def noDuplicatedMapName(name){
+    def names = ([] + c.openMindMaps*.name + c.openMindMaps*.root*.text).unique().sort()
+	println names
+    def i = 1
+    def nombre = name
+    while (names.contains(nombre)){
+        nombre = "${name} x${i++}".toString()
+    }
+    return nombre
+}
+
 // == MAIN ==
 def showIcons = true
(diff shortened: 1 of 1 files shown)


freeplane_plugin_script/scripts/apiGenerator.groovy
```diff
@@ -323,16 +389,52 @@ makeApi(proxy, Proxy.DependencyLookup.class)
...
+makeApi(utils, HtmlUtils.IndexPair.class)
 makeApi(utils, TextUtils.class)
 makeApi(utils, MenuUtils.class)
+makeApi(utils, MenuUtils.MenuEntry.class)
+makeApi(utils, MenuUtils.MenuEntryTreeBuilder.class)
+// org.freeplane.plugin.script
 makeApi(utils, FreeplaneScriptBaseClass.ConfigProperties.class)
+//org.freeplane.core.util
 makeApi(utils, FreeplaneVersion.class)
+//org.freeplane.launcher
```
dpolivaev, line 435: Are you sure we can add the Launcher? Why do you think we should?

EdoFro: I have no opinion on this.
I just added the classes that are shown in the API documentation. For example Launcher:

https://docs.freeplane.org/api/org/freeplane/launcher/Launcher.html

dpolivaev: Thanks, got it.

PR #1549, opened 2023-11-26
Copy style between maps should not copy attribute's formulas

EdoFro:
as said [here](https://github.com/freeplane/freeplane/pull/1539#issuecomment-1826904529)

I was thinking that adding formulas to attributes in template maps could be a way to add malicious code to many maps and users. That's why I added some code to neutralize any formula values in attributes copied as part of a style node


EdoFro:
now when copying a formula it adds a "_" at the left of the String:

example:

`= 5 + 8`  becomes `_= 5 + 8`

dpolivaev:
Is it a real problem? It adds complexity and might be no expected. What do you think about discussing it in the forum before developing the solution?

EdoFro:
I don't think it may be a real problem. I just realize it and I don't wanted to be the one that opened a breach.

Maybe it's better not to discuss this in the forum to not give away this idea.

EdoFro:
I just wanted you to know.

I made this PR, because I always try to propose a solution when I come with a problem.
I don't like persons that only see the problems and never try to help with any solution (even a bad one). 🤷🏼

dpolivaev:
I am hesitant to change things without discussion. 
Also, if somebody wanted to execute some code, it could be done using conditional styles and linked scripts.

So I suggest to close this PR.

EdoFro:
Ok. Good point. 👍🏼

PR #1691, opened 2024-02-13
adds System's user.country to defaultLocale

EdoFro:
Related to topic #1687

This change makes that the firstDayOfWeek in the Calendar panels is set to Sunday or Monday depending on the user's country.

For example, in my case ( System variable `user.country = CL` )

image

### DIFF
--- a/freeplane/src/main/java/org/freeplane/main/application/ApplicationResourceController.java
+++ b/freeplane/src/main/java/org/freeplane/main/application/ApplicationResourceController.java
@@ -131,7 +131,8 @@ public class ApplicationResourceController extends ResourceController {
 		}
 		resourceLoaders = new LinkedHashSet<>();
 		String country = (String) System.getProperties().get("user.country");
-		setDefaultLocale(props.getProperty(ResourceBundles.RESOURCE_LANGUAGE) + "_" + country);
+		country = country.length()>0?"_" + country:"";
+		setDefaultLocale(props.getProperty(ResourceBundles.RESOURCE_LANGUAGE) + country);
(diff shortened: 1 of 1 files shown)


dpolivaev:
It does not work like that because not all combinations of language and country are known to Java.
I am going to implement the fix myself, but for this, I need to understand what to do if the preferred combination is not available. I can either ignore the language or the country. If I ignore the language, standard dialogs would use the default language set by the OS. If I ignore the country, it would behave as before the fix.

EdoFro:
Hi Dimitry,
I thought that it was no problem, because when the string has a combination language/country, it splits them into two separate input parameters:

[look at line 341](https://github.com/EdoFro/freeplane/blob/0c44bb5b7f6db05fc90a80fd89b5048baf7cb986/freeplane/src/main/java/org/freeplane/main/application/ApplicationResourceController.java#L341)

```java
  localeDef = new Locale(lang.substring(0, 2), lang.substring(3, 5));
```

in my case I use english as language and live in Chile (where we speak spanish).

I use english just because its easier to post images and answer questions in the Discussion Forum.

I’m glad you had a better solution. I just tried to help. 😉👍

Best regards,

edo

EdoFro:
@dpolivaev,
I took a look at your commit and realized that a combination like I mentioned before (“en_CL”) is no possible. Just for you to know: I have no problem with that. My use case is a border case. I have no problem with my weeks starting on sunday or monday. It wasn’t an issue for me before. I mentioned it earlier just as an information, not as a request. 

BR

edo

PR #1741, opened 2024-03-16
Superuser -> superior

andrey-khropov:
I think that's what you really meant here, not running Freeplane as [a superuser](https://en.wikipedia.org/wiki/Superuser)


andrey-khropov:
Or maybe you meant [a power user](https://en.wikipedia.org/wiki/Power_user) ?

Issue #2499, opened 2025-06-07
Availability of ConfigProperties in script classes, e.g. installationBaseDir


HaaiHenkie:
**Describe the bug**
In a Groovy script, the Freeplane installation directory can be accessed using `config.installationBaseDir`. In a Groovy class, where the `config` binding is unavailable, one would expect to use `FreeplaneScriptBaseClass.ConfigProperties.installationBaseDir.` However, this results in an error stating the property doesn't exist — while simultaneously listing it as a "possible solution." Using the same code in a script produces the same error, suggesting the property is not statically accessible, contrary to what the message implies.

**To Reproduce**
1. Open Freeplane.
2. Create a new Groovy script node.
3. Paste and run the following code:
~~~groovy
import org.freeplane.plugin.script.FreeplaneScriptBaseClass.ConfigProperties
String installationDir = ConfigProperties.installationBaseDir
~~~
4. Observe the error:
~~~
No such property: installationBaseDir for class: org.freeplane.plugin.script.FreeplaneScriptBaseClass$ConfigProperties
Possible solutions: installationBaseDir
~~~

**Expected behavior**
* Either installationBaseDir should be accessible statically through ConfigProperties, or
* The documentation and error message should clearly state that installationBaseDir is not a static property and must be accessed differently.

At a minimum, the error message should not suggest a “possible solution” that is known not to work.

**Mind map files**
Not applicable — this is a scripting API issue.

**Screenshots**
Not applicable — the error is shown in the script execution console.

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Freeplane Version 1.12.11
 - Java Version 21

**Additional context**
As a workaround, the following code can be used in script classes to retrieve the installation directory using the script engine:

~~~groovy
import org.freeplane.plugin.script.proxy.ScriptUtils

def c = ScriptUtils.c()
String installationDir = c.script("config.installationBaseDir", "groovy")
~~~
However, this workaround is undocumented, and the current behavior remains misleading and problematic for users trying to write reusable code in classes.

This report has been generated by the Freeplane tutor in ChatGPT

macmarrum:
The method `org.freeplane.plugin.script.FreeplaneScriptBaseClass.ConfigProperties#getInstallationBaseDir` is public but not static. I think that one needs to instantiate ConfigProperties first:

```groovy
import org.freeplane.plugin.script.FreeplaneScriptBaseClass.ConfigProperties

def config = new ConfigProperties()
[...]

PR #2507, opened 2025-06-14
Javadoc improvement org.freeplane.plugin.script.FreeplaneScriptBaseClass

HaaiHenkie:
Fixed broken links in main description
In main description and ConfigProperties description: added reference to importing GroovyStaticImports in case of compiling source code outside Freeplane, see also https://github.com/freeplane/freeplane/issues/2499

### DIFF
--- a/freeplane_plugin_script/src/main/java/org/freeplane/plugin/script/FreeplaneScriptBaseClass.java
+++ b/freeplane_plugin_script/src/main/java/org/freeplane/plugin/script/FreeplaneScriptBaseClass.java
@@ -45,22 +45,30 @@ import groovy.lang.Script;
  * Only documented methods are meant to be used in scripts.
  * <p>The following global objects are provided as shortcuts by the binding of this class:
  * <ul>
- * <li><b>ui:</b> see {@link UITools}</li>
+ * <li><b>ui:</b> see {@link org.freeplane.core.ui.components.UITools UITools}</li>
  * <li><b>logger:</b> see {@link LogUtils}</li>
- * <li><b>htmlUtils:</b> see {@link HtmlUtils}</li>
- * <li><b>textUtils:</b> see {@link TextUtils}</li>
- * <li><b>menuUtils:</b> see {@link MenuUtils}</li>
+ * <li><b>htmlUtils:</b> see {@link org.freeplane.core.util.HtmlUtils HtmlUtils}</li>
+ * <li><b>textUtils:</b> see {@link org.freeplane.core.util.TextUtils TextUtils}</li>
+ * <li><b>menuUtils:</b> see {@link org.freeplane.core.util.MenuUtils MenuUtils}</li>
  * <li><b>config:</b> see {@link ConfigProperties}</li>
  * </ul>
  * The following classes may also be useful in scripting:
  * <ul>
- * <li>{@link FreeplaneVersion}</li>
+ * <li>{@link org.freeplane.core.util.FreeplaneVersion FreeplaneVersion}</li>
  * </ul>
  */
 public abstract class FreeplaneScriptBaseClass extends Script {
 	/**
-	 * Accessor for Freeplane's configuration: In scripts available
-	 * as "global variable" <code>config</code>.
+	 * <p>Accessor for Freeplane's configuration: In scripts available as "global variable" {@code config}</p>
+     *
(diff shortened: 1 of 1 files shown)


freeplane_plugin_script/src/main/java/org/freeplane/plugin/script/FreeplaneScriptBaseClass.java
```diff
@@ -45,22 +45,30 @@
...
- * <li>{@link FreeplaneVersion}</li>
+ * <li>{@link org.freeplane.core.util.FreeplaneVersion FreeplaneVersion}</li>
  * </ul>
  */
 public abstract class FreeplaneScriptBaseClass extends Script {
 	/**
-	 * Accessor for Freeplane's configuration: In scripts available
-	 * as "global variable" <code>config</code>.
+	 * <p>Accessor for Freeplane's configuration: In scripts available as "global variable" {@code config}</p>
+     *
```
dpolivaev: I do not understand what you mean by "utility scripts and addon classes".
If I understand it correctly, the variable is available in all classes compiled by Freeplane, but it is not available in classes compiled outside. Am I right?

HaaiHenkie: I mean that the global variable `config` is only available in scripts. This variable (like `node` and `c`) is **not** available in classes compiled by Freeplane, i.e. [utility scripts](https://docs.freeplane.org/scripting/Your_own_utility_script_library.html) and classes compiled to jar when packaging an add-on.

dpolivaev: I have just commented at https://github.com/freeplane/freeplane/issues/2499#issuecomment-2972700333

HaaiHenkie: So you were right, it is not available when compiling groovy classes outside Freeplane. I adjusted the javadoc accordingly.

dpolivaev: I added another comment here: https://github.com/freeplane/freeplane/issues/2499#issuecomment-2973635300.
Could you check it and see if it makes sense to include it in the documentation?

And I want to highlight something important: it was your commitment — shown in this PR — that made me dive into the code and try to understand what’s really going on.
Your contribution already made a difference, and I truly hope you’ll keep going.

HaaiHenkie: In the main description of `FreeplaneScriptBaseClass` and in the `ConfigProperties` description, I have added a reference to importing `GroovyStaticImports` in case of compiling source code outside Freeplane.

I looked into `GroovyStaticImports`. I want to make sure if I understand it well and I would like to make some suggestions in the end.

It seems that `GroovyStaticImports` is the class that is injecting the DSL shortcuts for the 6 global classes available in the Freeplane scripting environment into any utility script or groovy source code on the script classpath. Since it is dated 30 January 2024, it means that before that time you needed to add this imports manually to your groovy source code. That is what I had in mind writing the first PR.

So `GroovyStaticImports` imports the same 6 global classes as `FreeplaneScriptBaseClass`.

In addition `GroovyStaticImports` imports `ScriptUtils` as `scriptUtils` and `ScriptUtils.ignoreCycles()` as `ignoreCycles()`. As far as I know this is not documented so I doubt if anybody is using this imports in utility scripts.

`GroovyStaticImports` does not import the static methods that `FreeplaneScriptBaseClass` provides without prefix, like `N(String id)`, `T(String id)`, `V(String id)` and `ifNull(Object value, Object valueIfNull)`. I don't mind since a hardly ever use these methods. But if someone is using this in a utility script it will give an error.

My first recommendation is to document `GroovyStaticImports`.

My second recommendation is to add `GroovyStaticImports` to the Freeplane API and to remove it as an automatic injection. This has the following advantages:
1. It can be documented as part of the API
2. Globals `scriptUtils` and `ignoreCycles()` are recognised in the IDE
3. If the source code is a groovy class, all other globals are recognised in the IDE (since in this situation their definition is not provided by `freeplane.dsld` or `freeplane.gdsl`)
4. There is no difference between compiling utility scripts or source code inside or outside Freeplane

PR #2756, opened 2025-11-23
Replace DES with AES-256 (again)

oyvinrog:
This is the continuation of another pull request that was reverted (73858dd).

The other pull request did actually work when creating **new** maps or nodes, but it did not overwrite legacy DES maps with the new AES-256 scheme. That was based on a misunderstanding. 

I have now updated the code so that when a user unlocks a legacy DES-encrypted node and re-locks it, the node is re-encrypted with AES-256 instead. This means BOTH new encryptions and legacy maps will automatically migrate to AES-256 through normal use (unlock/lock cycles).

I have done a series of tests and verifications. i.e. edit legacy encrypted maps to see whether they are overwritten with AES-256, and edit new maps. When checking the mm files, AES-256 (64 hex characters) appear as expected
The automatical tests also execute successfully.

The solution is in the following 4 lines in EncryptionController.java:

```
if (decrypted) {
            final IEncrypter aesEncrypter = EncryptionHelper.createEncrypter(password);
            encryptionModel.setEncrypter(aesEncrypter);
            tempEncrypter.destroy();
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -261,3 +261,7 @@ gradle-app.setting
 *.backup
 *.temp
 .!*
+
+# Local testing scripts
+test_encryption_quick.sh
+.classpath_cache
--- a/freeplane/src/main/java/org/freeplane/features/encrypt/Aes256Encrypter.java
+++ b/freeplane/src/main/java/org/freeplane/features/encrypt/Aes256Encrypter.java
@@ -1,0 +1,211 @@
+/*
+ *  Freeplane - mind map editor
+ *  Copyright (C) 2025 Freeplane team
+ *
+ *  This program is free software: you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
+ *  the Free Software Foundation, either version 2 of the License, or
+ *  (at your option) any later version.
+ *
+ *  This program is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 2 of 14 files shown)


dpolivaev:
I think that overwriting makes sense because there is no easy way to replace the encryption otherwise.

dpolivaev:
I think there might be a flaky test because I got

````

Aes256EncrypterTest > slightlyWrongPasswordReturnsNull FAILED
    java.lang.AssertionError: 
    Expected: null
         but: was "��U�@2���=�Ji"
        at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:20)
        at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:8)
        at org.freeplane.features.encrypt.Aes256EncrypterTest.slightlyWrongPasswordReturnsNull(Aes256EncrypterTest.java:278)
````

https://github.com/freeplane/freeplane/actions/runs/19617132161/job/56171379363

dpolivaev:
Also I have noticed that you use hamcrest assertions in your tests. Actually I prefer AssertJ. I think AI can easily adjust the tests. I would appreciate it if you could open a new PR.

dpolivaev:
Also other WrongPassword tests turned out to be flaky, I added `@Ignore` annotation to them so that they do not block my builds.

dpolivaev:
https://github.com/freeplane/freeplane/releases/tag/release-1.12.15-pre18

PR #2758, opened 2025-11-23
Fix problematic test by switching to AES-GCM

oyvinrog:
AES-CBC occasionally decrypts with wrong password when garbage output has valid-looking padding (~1/256 chance). GCM provides authenticated encryption that always fails with AEADBadTagException on wrong password.

Changes:
- IV: From 16 to 12 bytes (GCM standard)
- Cipher: From AES/CBC/PKCS5Padding to AES/GCM/NoPadding
- Spec: From IvParameterSpec to GCMParameterSpec(128)

No compatibility impact - feature added Nov 23, not yet in stable release.
Important: Note that maps created with the version in between cannot be opened with the newer code

I now use AssertJ  instead of hamcrest

did a full gradle test and encryption test. Everything passed. Also did manual tests.

### DIFF
--- a/freeplane/src/main/java/org/freeplane/features/encrypt/Aes256Encrypter.java
+++ b/freeplane/src/main/java/org/freeplane/features/encrypt/Aes256Encrypter.java
@@ -26,13 +26,14 @@ import java.security.spec.InvalidKeySpecException;
 import java.security.spec.KeySpec;
 import java.util.Arrays;
 
+import javax.crypto.AEADBadTagException;
 import javax.crypto.BadPaddingException;
 import javax.crypto.Cipher;
 import javax.crypto.IllegalBlockSizeException;
 import javax.crypto.NoSuchPaddingException;
 import javax.crypto.SecretKey;
 import javax.crypto.SecretKeyFactory;
-import javax.crypto.spec.IvParameterSpec;
+import javax.crypto.spec.GCMParameterSpec;
 import javax.crypto.spec.PBEKeySpec;
 import javax.crypto.spec.SecretKeySpec;
 
@@ -41,10 +42,10 @@ import org.freeplane.features.map.IEncrypter;
 
 public class Aes256Encrypter implements IEncrypter {
 	private static final int SALT_LENGTH = 16;
-	private static final int IV_LENGTH = 16;
+	private static final int IV_LENGTH = 12;
 	private static final int KEY_LENGTH = 256;
 	private static final String KEY_DERIVATION_ALGORITHM = "PBKDF2WithHmacSHA256";
-	private static final String CIPHER_ALGORITHM = "AES/CBC/PKCS5Padding";
+	private static final String CIPHER_ALGORITHM = "AES/GCM/NoPadding";
 	private static final int ITERATION_COUNT = 100000;
 	
 	private Cipher dcipher;
@@ -94,6 +95,8 @@ public class Aes256Encrypter implements IEncrypter {
 			}
 			final byte[] utf8 = dcipher.doFinal(ciphertext);
 			return new String(utf8, StandardCharsets.UTF_8);
+		}
+		catch (final AEADBadTagException e) {
 		}
 		catch (final BadPaddingException e) {
 		}
@@ -162,14 +165,14 @@ public class Aes256Encrypter implements IEncrypter {
 				if (iv == null) {
 					currentIV = new byte[IV_LENGTH];
 					secureRandom.nextBytes(currentIV);
-					final IvParameterSpec ivSpec = new IvParameterSpec(currentIV);
+					final GCMParameterSpec gcmSpec = new GCMParameterSpec(128, currentIV);
 					ecipher = Cipher.getInstance(CIPHER_ALGORITHM);
-					ecipher.init(Cipher.ENCRYPT_MODE, key, ivSpec);
+					ecipher.init(Cipher.ENCRYPT_MODE, key, gcmSpec);
 				} else {
 					currentIV = iv;
-					final IvParameterSpec ivSpec = new IvParameterSpec(currentIV);
+					final GCMParameterSpec gcmSpec = new GCMParameterSpec(128, currentIV);
 					dcipher = Cipher.getInstance(CIPHER_ALGORITHM);
-					dcipher.init(Cipher.DECRYPT_MODE, key, ivSpec);
+					dcipher.init(Cipher.DECRYPT_MODE, key, gcmSpec);
 				}
 			}
 			catch (final InvalidAlgorithmParameterException e) {
--- a/freeplane/src/test/java/org/freeplane/features/encrypt/Aes256EncrypterTest.java
+++ b/freeplane/src/test/java/org/freeplane/features/encrypt/Aes256EncrypterTest.java
@@ -17,24 +17,12 @@
  */
 package org.freeplane.features.encrypt;
 
-import static org.hamcrest.MatcherAssert.assertThat;
-import static org.hamcrest.Matchers.equalTo;
-import static org.hamcrest.Matchers.not;
-import static org.hamcrest.Matchers.notNullValue;
-import static org.hamcrest.Matchers.nullValue;
-import static org.junit.Assert.assertNotEquals;
-import static org.junit.Assert.assertTrue;
+import static org.assertj.core.api.Assertions.assertThat;
 
 import org.freeplane.features.map.IEncrypter;
 import org.junit.After;
 import org.junit.Test;
 
-/**
- * Specific tests for AES-256 encryption implementation.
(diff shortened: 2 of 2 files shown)


oyvinrog:
With AES-CBC mode (that was used in the previous PR):

When you use the wrong password, the decryption process still runs and produces complete garbage data
But here's the catch: CBC mode requires data to have special "padding" bytes at the end.

The system checks: "Does this look like valid padding?" With random garbage, there's about a 1 in 256 chance the garbage accidentally looks like valid padding

For example, if the last byte happens to be 01, that's valid padding!  When this happens, the decryption "succeeds" but gives you nonsense output instead of an error


GCM is like having a tamper-evident seal in addition to the lock It creates a cryptographic "signature" (authentication tag) that proves:

1) The right password was used

2) The data hasn't been corrupted

With the wrong password, this signature check always fails immediately with AEADBadTagException

**Caution:**

The AES-GCM implementation created on November 23, 2025 is **NOT backward compatible** 
with the AES-CBC implementation also created on November 23, 2025.

- **Maps encrypted with AES-GCM** cannot be decrypted by the AES-CBC version
- **Maps encrypted with AES-CBC** cannot be decrypted by the AES-GCM version
- Both versions remain compatible with **older DES-encrypted maps**

dpolivaev:
I’ve merged the changes, although a few issues remain unresolved: some test cases still don’t use AssertJ assertions, and several tests marked with `@Ignore` are still disabled and haven’t been reworked.

I want to mention this openly, because the current state is not fully aligned with what we discussed. I would appreciate a follow-up to address these points.

dpolivaev:
On the other side, maybe there are too many tests? We should not test the libraries, only our own logic.

oyvinrog:
Hi, I agree 

I propose to:

Convert all tests to AssertJ (EncryptionHelperTest.java and EncryptionTest.java)

Remove the 4 @Ignore tests. All 4 test "wrong password returns null" scenarios. They test cryptographic authentication (a library feature, not our logic). We already have one passing test (Aes256EncrypterTest.wrongPasswordReturnsNull) that validates our error handling returns null, so these are redundant.

Remove ~23-25 tests that test library behavior rather than Freeplane logic:
Tests validating UTF-8 encoding (unicode, emoji, long text, special characters, XML)
Tests validating PBKDF2 password handling (short/long/unicode/special char passwords)
Duplicate or redundant test cases

This will reduce the test count by ~40% while keeping full coverage of our actual logic: version markers, salt/IV generation, data structure format, error handling, algorithm detection, and backward compatibility.

Would you like me to submit a PR with these changes?

dpolivaev:
Yes, sure.

Unfortunately, I am currently investigating a case where one node of my test map can not be decrypted. In old versions of the same map I could decrypt it (with AES256), and I do not remember changing the node password. Maybe I did, maybe not. I can not tell it for certain, so all we can is observe it before I release it as stable. We should be really sure no mind maps can become corrupted.

Also I wonder why you create temp decrypter and replace it even if the algorithm was already AES

````
        final IEncrypter tempEncrypter = EncryptionHelper.createDecrypter(password, encryptedContent);

        boolean decrypted = encryptionModel.decrypt(mapController, tempEncrypter);

        if (decrypted) {
            final IEncrypter aesEncrypter = EncryptionHelper.createEncrypter(password);
            encryptionModel.setEncrypter(aesEncrypter);
            tempEncrypter.destroy();
        } else {
            tempEncrypter.destroy();
        }
````

**Comment by oyvinrog** [comment]
if you try to decrypt a map from the previous PR with AES CBC with the current PR with AES GCM, then that is expected behavior (as mentioned in the previous PR).

if you try to decrypt a map that used to be DES with AES CBC or GCM, then thats strange (I tried that manually and it worked. Also I believe test like createDecrypterFallsBackToSingleDesForLegacyContent covers this). Do you happen to have a copy of your test map? I can probably check which algorithm is used.

I will try it again on my side (decrypt a legacy map in the new version).


We can do the following, to make it simpler:

final IEncrypter encrypter = EncryptionHelper.createDecrypter(password, encryptedContent);
boolean decrypted = encryptionModel.decrypt(mapController, encrypter);

if (decrypted) {
    if (encrypter instanceof Aes256Encrypter) {
        encryptionModel.setEncrypter(encrypter);  // Reuse
    } else {
        // Legacy DES: upgrade to AES-256
        final IEncrypter aesEncrypter = EncryptionHelper.createEncrypter(password);
        encryptionModel.setEncrypter(aesEncrypter);
        encrypter.destroy();
    }
} else {
    encrypter.destroy();
}

**Comment by dpolivaev** [comment]
The backup version of the map can be decrypted. The both encrypted contents start with the correct prefix.

Good:

```
<node TEXT="eee" ENCRYPTED_CONTENT="FP-AES256-V1:1iJ08Fg2HE+LhjkQa1HH7lap69aJ62Yf5ljogm/PY2vKie9nCaEEkRBtolpnSj1+Kjn7BZjSctrLvI8lyog1Wa39olnUPA+FLn9H6AaKIEBa0R668bvrqqbAC2Rifjy9gQ9R9zJfUjvYozr8ZZC5tM4ywgCPg4BE3mbWqie4r2V94nTp2K474fHmgTeyI796XiCkZjbavropXxiskPIGOHg82P2HusljYoF5G2+LPioSovWsDajuCE+hvvh5Ujw4jyNJOiYtrAdQkgI78OsvrQdGtPCxTZfiCZXJUuXnRfpl0SulYZ5tGf+cF1tyxRTIc7WBz6LBO2hIdClrAw2Q/kSAKf9JHFqeCs3AMqDspt0TQRzBcPfRwP2pTMX2Lf2VR91ZMHp+fETrQ2bV3SK2TAuxxyOKD/UgjiwY0RdfaPBCs8lR3yyXkiLRaybCGkw//jmc" ID="ID_1358000801"/>
```

Bad:

```
<node TEXT="eee" ENCRYPTED_CONTENT="FP-AES256-V1:3zqgepk/zUq+K91duTadotfevv7G7IPVB/hTjEbNH94GJHYvLAo4jKWLDr98bEmbhodqRuzjvv4uHs7yYl+zlbPxSAXcur/g4qZH0xNSdFCh35snJLshQa/F1LBKyLPjffofm1yIdfz3oatXsDA8QakFvvG5VirJKBLFrFlebHM3e6iAzSgeFF1uEhZXEMhOvDpCnuPjKLiPixqhu6XTI0oBfMqX7/ugTstf62ua+tE=" ID="ID_1358000801" VSHIFT_QUANTITY="33.23077 pt"/>
```

dpolivaev:
The password was `loser3` :)

dpolivaev:
Related to reuse, I would reuse the encrypter if the encrypted context starts with the correct (current default) prefix.

dpolivaev:
By the way, the encryption / decryption code is one of the worst and less logical parts of Freeplane. Maybe we should clean it up in a remote pairing session. How do you see it? I was hesitant to touch it because it is not only ugly but also many people rely on it and no recovery is possible if things go wrong.

oyvinrog:
my analysis of your provided ciphers tells me that you used AES CBC to decrypt GCM, You cannot use GCM to decrypt CBC or vice versa, as mentioned in the previous PR.


I am able to successfully decrypt your "bad" cipher with CBC aES 


--- Trying GCM structure (12-byte IV) ---
IV (bytes 16-28): d7debefec6ec83d507f8538c
Ciphertext+Tag length: 148 bytes
GCM Decryption: FAILED: InvalidTag: 

--- Trying CBC structure (16-byte IV) ---
IV (bytes 16-32): d7debefec6ec83d507f8538c46cd1fde
Ciphertext length: 144 bytes
CBC Decryption: `<node TEXT="fff" ID="ID_1185647756">
<node TEXT="ggg" FOLDED="true" ID="ID_1263449646">
<node TEXT="hhh" ID="ID_1367648672"/>
</node>
</node>`

oyvinrog:
Tried myself with success:

- create map with old version (DES). Decrypted successfully and saved as AES successfully
- create map with new version (AES). Decrypted successfully

as mentioned above. Managed to successfully decrypt your "bad" example with AES CBC (cannot use the new PR to decrypt the old PR, as mentioned previously).

(a) Newest version: AES GCM
(b) Version that we had on Sunday: AES CBC
(c) Previous production version: DES

(a) will decrypt (c) successfully. (b) will decrypt (c). BUt (a) can not be used to decrypt (b) or vice versa.

oyvinrog:
> By the way, the encryption / decryption code is one of the worst and less logical parts of Freeplane. Maybe we should clean it up in a remote pairing session. How do you see it? I was hesitant to touch it because it is not only ugly but also many people rely on it and no recovery is possible if things go wrong.

I like this, but I am heavily occupied between coding sessions. So I prefer submitting and refining PR's. I have no problem if you want to roll this back, and I will try to create an even more robust code.

PR #2762, opened 2025-11-29
Feature/migrate encryption tests to assertj 

oyvinrog:
1) Migrate rest of the encryption test assertions from Hamcrest to AssertJ.

2) Remove `@Ignore` annotations from tests that are now passing with GCM encryption (introduced in recent commits). All 107 encryption tests now succeed.

Note that I will propose another PR to simplify and remove tests once this is good (or if you want, I can add some more commits here). 

Changes Made:

- **Migrated 4 test files** from Hamcrest to AssertJ assertions:
  - `Aes256EncrypterTest.java`
  - `EncryptionHelperTest.java`
  - `EncryptionModelTest.java`
  - `EncryptionTest.java`

NOTE: The tests pass on my side, but I am not 100 % sure that CBC was the reason these were failing on your side. Appreciate if you could check again.

Testing:

- All 107 encryption tests pass successfully
- Tests that were previously failing with CBC encryption now work correctly with the upgraded GCM encryption mode
- Manual test was conducted again, by decrypting legacy map and decrypting a new map. Both tests were good.

### DIFF
--- a/freeplane/src/test/java/org/freeplane/features/encrypt/Aes256EncrypterTest.java
+++ b/freeplane/src/test/java/org/freeplane/features/encrypt/Aes256EncrypterTest.java
@@ -248,7 +248,6 @@ public class Aes256EncrypterTest {
 	}
 
 	@Test
-	@Ignore
 	public void slightlyWrongPasswordReturnsNull() {
 		final StringBuilder password1 = new StringBuilder("password");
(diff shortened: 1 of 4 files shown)


oyvinrog:
just some more background: Previous tests were (MOST LIKELY) failing because of a known issue with the AES CBC mode. Because we now use GCM, they should not have that particular issue.  They all passed on my side after doing this update

PR #2764, opened 2025-11-30
Feature/encryption quality improvements

oyvinrog:
This PR improves the encryption module's code quality and test reliability with resource management enhancements, better documentation, and test cleanup.

- **Fixed DES decryption test**: Properly handles the fact that DES encryption (lacking authentication) may return either `null` OR garbage data on wrong password, but should never return correct plaintext

- **Removed duplicate tests**: Cleaned up redundant test cases to reduce maintenance burden

- **Added try-finally blocks**: Guarantee that decrypters are always properly destroyed after use, preventing potential resource leaks

- Documented why fresh encrypter instances are created in `EncryptionController`, improving code understanding for future maintainers

Testing:

- All encryption tests pass consistently
- Tests cover both legacy (DES) and modern (AES-GCM) encryption paths
- Manual tests passed 

Changes:
- `freeplane/src/test/java/.../controller/encryption/` - test files
- `freeplane/src/main/java/.../controller/encryption/EncryptionController.java` - try-finally blocks and comments only

No changes to core encryption logic or public APIs.

### DIFF
--- a/freeplane/src/main/java/org/freeplane/features/encrypt/EncryptionController.java
+++ b/freeplane/src/main/java/org/freeplane/features/encrypt/EncryptionController.java
@@ -151,19 +151,25 @@ public class EncryptionController implements IExtension {
     private boolean decrypt(final EncryptionModel encryptionModel, final StringBuilder password) {
         final MapController mapController = Controller.getCurrentModeController().getMapController();
         final String encryptedContent = encryptionModel.getEncryptedContent();
-        final IEncrypter tempEncrypter = EncryptionHelper.createDecrypter(password, encryptedContent);
+        final IEncrypter encrypter = EncryptionHelper.createDecrypter(password, encryptedContent);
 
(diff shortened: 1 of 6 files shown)


oyvinrog:
notice the comment in 00688d3. I know you don't like comment clutter very much, but I can't see any better way to communicate that the code should not be changed. If not creating a new encrypter for every call to decrypt(), decryption will fail (each new encryption must use new random values).

dpolivaev:
Thank you for the pull request. I think the current Encrypter design could be simplified by switching to a more functional, data-flow-oriented structure and eliminating member fields entirely. All intermediate values have very short lifetimes, and making the data flow explicit would reduce accidental complexity and lower the chance of subtle bugs. You could ask an AI assistant to rewrite the code accordingly.

What is your view on restructuring it in this direction?

Since the original AES implementation has already been merged, I can integrate further improvements and refactorings once the accidental complexity in this area is reduced.

Issue #2941, opened 2026-07-21
New child node is invisible when created inside a jumped-in node


euu2021:
When I'm jumped into a node (that node is the root of what I'm viewing) and I create a new child node under it, nothing appears on screen. The node really is created — if I jump out, it's there in the map — but while I stay inside, I see nothing, as if the command did nothing.

**How to see it**

1. In Preferences → Environment → Load, set folding to "always fold all after load".
2. Open a map, jump into a node a few levels deep, and close Freeplane.
3. Reopen it — Freeplane restores you inside that node.
4. Inside, pick one of the children and jump into it too.
5. Create a new child node under it (Insert → New Child Node).

Expected: the new child appears right away.
Actual: nothing appears. Only after jumping out do I see that the node was created.

I've opened a pull request with a fix.

dpolivaev:
https://github.com/freeplane/freeplane/releases/tag/release-1.13.4-pre02

PR #2943, opened 2026-07-21
Keep the view root registered as a viewer of its node

euu2021:
Fixes 2941.

## What happens

When a child is created on a node that is the jumped-in view root, the child is added to the model but no view is drawn for it; it appears only after a later jump out.

When a node becomes the view root, the root swap tears down the old root subtree with `NodeView.remove()`, which recurses into child views. If the node being jumped into was a child of the old root, that recursion deregisters its `NodeView` as a viewer of its `NodeModel`, and nothing re-registers it. So while the node is the view root, `NodeModel.fireNodeInserted` — which iterates the node's viewers — never reaches the map root view (only the outline's viewer remains). The inserted child updates the model but gets no view until a later jump out rebuilds the subtree.

## Change

Re-register the current root view as a viewer of its node after the swap in `MapView.setRootNode(NodeView, RootChange)`, guarded with `getViewers().contains` so a node that is already a viewer is not added twice.

## Verification

- No automated regression: reproducing it needs the reuse jump-in path (a node with an already-built child view before becoming the view root), which the headless test harness cannot set up because child-view creation is deferred until the map is displayable. The closest automated substitute would need a real displayable `MapView`.
- Verified manually with a deterministic reproducer: jump into a node that is the restored view root, create a child — with the change it appears immediately, and jump out stays consistent.
- The `:freeplane:test` suite passes.

Issue: https://github.com/freeplane/freeplane/issues/2941

### DIFF
--- a/freeplane/src/main/java/org/freeplane/view/swing/map/MapView.java
+++ b/freeplane/src/main/java/org/freeplane/view/swing/map/MapView.java
@@ -3343,6 +3343,8 @@ public class MapView extends JPanel implements Printable, Autoscroll, IMapChange
 		}
 		else
 		    rootsHistory.clear();
+		if(! newRootView.getNode().getViewers().contains(newRootView))
+		    newRootView.getNode().addViewer(newRootView);
 		if(nextSelectedNode.getParent() == null || ! nextSelectedNode.isContentVisible())
 		    nextSelectedNode = newRootView;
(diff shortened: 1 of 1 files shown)


dpolivaev:
I think it does not solve the issue completely.

dpolivaev:
I am working on another fix

PR #2963, opened 2026-08-05
Notes have tabs feature with links within

mayankdmishra:
This PR adds tabs to the notes panel. So a user can now have multiple tabs for more categorised info for the node. Also, the text within the tab, can hyperlink to the nodes or any noteTab for quick access

### DIFF
--- a/freeplane/src/main/java/org/freeplane/features/note/NoteBuilder.java
+++ b/freeplane/src/main/java/org/freeplane/features/note/NoteBuilder.java
@@ -55,14 +55,23 @@ class NoteBuilder implements IElementContentHandler {
 	        final String text = content != null ? content.trim() : null;
 	        final Object typeAttribute = element.getAttribute(NodeTextBuilder.XML_RICHCONTENT_TYPE_ATTRIBUTE, null);
 			if (NodeTextBuilder.XML_RICHCONTENT_TYPE_NOTE.equals(typeAttribute)) {
-				final NoteModel note = new NoteModel();
-				note.setXml(text);
-	            final String contentType = element.getAttribute(
-	                    NodeTextBuilder.XML_RICHCONTENT_CONTENT_TYPE_ATTRIBUTE,
-	                    ContentSyntax.XML.prefix);
-	            note.setContentType(ContentSyntax.specificType(contentType));
-
-				((NodeModel) node).addExtension(note);
+				final NodeModel nodeModel = (NodeModel) node;
+				final String tabName = element.getAttribute("TAB", null);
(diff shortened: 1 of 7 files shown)


macmarrum:
Hi @mayankdmishra 

Thanks for the PR. The idea is interesting. 

I think this would be a big change; it would be beneficial for you to describe the problems which are solved by introducing Note Tabs. In other words, how you (intend to) use Note Tabs.

Let's also keep in mind that Freeplane has a lot of functionality linked to Note. Here's a high-level analysis: https://chat.z.ai/s/137fd6cb-3bbf-4887-a787-193b464d1203

dpolivaev:
Hi  @mayankdmishra,

thank you very much for your submission. If I see it correctly you want to allow more than one note per node so that they can be named and shown in tabs. I am not sure why exactly you need this feature and I am generally hesitant to add new features because Freeplane is quite a complex application right now and also because the new feature would also influence exports and require a format change. Also I do not think that "tabs" is a right term, but this discussion would only make sense if your idea is accepted as a feature request.

You can start a discussion and present your idea to the community. If it gets support I would consider how to implement it right.

Regards

dpolivaev:
Closed because of no contributor response

mayankdmishra:
@macmarrum my bad, for not providing the description here or not starting a discussion as suggested by @dpolivaev. Both tasks kept postponing for some other projects/schedules. But I am here now prioritizing it. Here is the description as to why I needed (and why others may need) such features.

1. For me, categorisation matters, the more simple, the better. So e.g. an idea/program has class/category and a class may have some versions. So such class/func versions can be put in their respective tabs (of same node of class/idea). Also, more that one tab (of course @dpolivaev said that tabs may not be the right term, so I have removed the 'tab' term in my commit and is used here just to refer the theoretical concept) opens door to jot down more categorised ideas (for same Node) without cluttering the same tab. 

2. The text inside the Note can be selected and with right-click context-menu it can hyperlink to any other Note{Tab} available under any Node or simply to any Node, for quick navigation and exploration.

3. I believe that ideas are precious and just because of a decisive moment, that idea should not be gone permanently from the map. Who knows when it is needed later. Thus, the deletion of Note {tab} doesn't directly delete it. As a safety valve a Trash node is created automatically, to contain the deleted Note {tab}. This Trash node is immutable so nothing can be added to it. There are only two possibilities either restore the node a user wants or Delete the whole chain of node. Yes deleting a single node from Trash is not allowed to keep the chain intact, as is. Same deletion feature (of immutable Trashing) is implemented to the Nodes as well, for safety valve.

4. As @dpolivaev has suggested I will open a discussion but it may take some time, a day or two. Please bear with me. In that discussion I will also give some screenshots of what I have got at my end just to give a feel of the features.

5. Also, as this PR is closed, do I have to create another PR or the same can be reopened?

Regards

PR #2982, opened 2026-09-07
Fix popup placement on multi-monitor setups (FlatPopupFactory retention and X11 coordinate resynchronization)

hyperdreamer:
## Problem

Freeplane popups (menu bar popups, combo-box drop-downs, context menus, tooltips) are misplaced on multi-monitor setups with mixed resolutions/scales, and on Linux/X11 (Java 11 + KWin) the whole frame's coordinate cache can desync from the window manager so popups appear on the wrong monitor. Two distinct defects:

1. `FrameController` unconditionally replaced FlatLaf's multi-monitor-aware `FlatPopupFactory` with Swing's basic `PopupFactory`, which recycles heavyweight popup windows without checking their `GraphicsConfiguration`, so recycled popups render on the monitor/scale of the window they were first created on.
2. On OpenJDK 11 under X11 with KWin, AWT's top-level window position cache does not update (fully or partially) when the window manager places, maximizes, restores, or moves the window, while the visible native window is elsewhere. Popups anchor to the stale AWT coordinates.

## Changes

**Cross-platform popup factory (all OS)**
- `FrameController`: stop discarding `FlatPopupFactory`; add `configurePopupFactory()` invoked from `fixLookAndFeelUI()` so it is authoritative on startup and on every look-and-feel change.
- New preference `freeplane.popup_factory=auto|flatlaf|basic` (default `auto`) with an explicit fallback for users.
- `Compat.isJetBrainsRuntime()` detection: on JetBrains Runtime, `auto` disables `Popup.dropShadowPainted` (the original reason the factory had been disabled); `basic` remains an escape hatch.

**X11 frame coordinate resynchronization (X11 toolkit only)**
- `ApplicationViewController.FrameResynchronizer`: resynchronize the frame with the window manager geometry on open, maximize transitions, unmaximize restore, and normal-state move/resize events.
- Native geometry is derived from the X11 shell window (`XTranslateCoordinates` + `XGetGeometry`) plus `_NET_FRAME_EXTENTS` via the already bundled JNA (`jna-platform`), converted to logical pixels and guarded by a plausibility check.
- The native position is only adopted when window sizes agree (transient window-manager geometry is never applied); a bounded deferred recheck (10 x 200 ms) covers delayed AWT configure notifications and window-manager animations.
- All native handling is gated on `sun.awt.X11.XToolkit`; Windows, macOS and native-Wayland toolkits are unaffected.

## Testing

- New unit tests: `ApplicationViewControllerTest` (screen-target selection, maximize/unmaximize/normal-window resync conditions — including the stale-position-versus-native-size cases), `PopupFactoryConfigurationTest` (factory selection, L&F transitions, preference overrides), `CompatTest` (JetBrains Runtime detection).
- `gradle :freeplane:test` passes (all suites).
- Manually verified on a 3-monitor X11/KWin setup (1920x1080 + 3200x1800 + 1920x1080, Java 11): startup-maximized on a secondary monitor, cross-monitor moves, maximize/restore cycles, and popup placement all keep AWT coordinates equal to the native window geometry (popup anchor == visible window).

### DIFF
--- a/freeplane/src/main/java/org/freeplane/core/util/Compat.java
+++ b/freeplane/src/main/java/org/freeplane/core/util/Compat.java
@@ -10,6 +10,7 @@ import java.net.URI;
 import java.net.URL;
 import java.util.Arrays;
 import java.util.HashSet;
+import java.util.Locale;
 import java.util.Set;
 
 import javax.swing.JFrame;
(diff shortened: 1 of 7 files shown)


hyperdreamer:
Closing this combined PR so the changes can be reviewed as separate focused PRs: one for FlatLaf popup-factory retention and one for X11 frame-coordinate resynchronization. No replacement PR is being submitted yet.

Issue #1037, opened 2023-03-04
Relative paths for images


linyangqi:
I found that images are saved as absolute paths, but I really wish it could be optionally relative paths.
Because the absolute paths are different between Windows and Linux: 
![图片](https://user-images.githubusercontent.com/47072330/222901161-b2231d7c-99ae-4507-8c20-921afc184fc0.png)

Furthermore, relative paths are also more friendly for the .mm to share on Internet (such as github) - if the images are saved as  absolute paths, those recieved .mm files have to fix paths before reading.

SceneryMC:
Freeplane supports relative paths. You can open the `.mm` file with a text editor such as VS Code, then replace`file:///media/WBW/杂文件/大学/【学期分类】/大三下/理论力学B/` with `./`
After that, change `Links` to `Relative` in `Tools - Preferences - Environment - Hyperlink types`

freeplane支持相对路径。你可以用VS Code等文本编辑器打开`.mm`文件，将`file:///media/WBW/杂文件/大学/【学期分类】/大三下/理论力学B/`替换为`./`
之后，在`工具-首选项-环境-超链接类型`中将`链接`改为`相对`即可

linyangqi:
@SceneryMC Thank you for the help! Yet I find it if I directly add picture by ctrl+v, the picture is in relative url, but if I add picture manually with "insert-image- add image"(?) the picture will be saved as absolute url: 
![image](https://user-images.githubusercontent.com/47072330/235350275-4bd0e9c2-4175-48ce-9d0d-d3a6e4e98e97.png)

![image](https://user-images.githubusercontent.com/47072330/235350091-78915dca-49ed-4a12-bb4d-c28ae05e20e0.png)

(the first is manually added pic, the second si by ctrl+v)


By the way I wonder how can I change the image pasting option so that it won't always create "xxx_files" folder XD.

Issue #1019, opened 2023-02-20
Edit Hyperlink - Removed - Not included.


bstott:
Windows 10
Freeplane 1.10.6

**Describe the bug**
No Edit Hyperlink or Modify Hyperlink Feature is available. 

**To Reproduce**
1. Go to Node
2. Right Click on Node.
3. Go to Edit.  
4. There is no selection to Edit Hyperlink.

OR

1. Go to Tool Bar > Edit...
2.  Nothing exists for Hyperlink.

OR

1. Go to Edit Menu
3. Nothing Exists regarding Hyperlinks. 

There is also no Edit/Modify under Insert Hyperlink.  But, strangely you have 'Remove." Which should be under Edit?  Why Remove when you go to Insert?

How could you miss, omit, remove, and not have Edit Hyperlink on Freeplane ???

Please fix. 

**Expected behavior**
Expect to have easy capability to Edit any and ALL of anything I type into the software.
Expect an easy Right Click > Edit > Hyperlink. And to have Edit/Modify Hyperlink under Edit... in toolbar, under Edit Menu, and anywhere else you include Node Edit/Modifications.

**Screenshots**
Can not get screenshot of mouse over menus with a Snip program and  Window+PrntScrn  isn't effective with multiple monitors. The text is not legible. So, I did not include.  

**Desktop (please complete the following information):**
 - OS: [e.g. iOS]   Windows 10
 - Version [e.g. 22]  1909
 - Java Version   19 2022-09-20

**Additional context**
This is frustrating.

theelectricsoul:
You'll find those commands through the top menu bar or when making a right click on the node.
![image](https://user-images.githubusercontent.com/10260832/221162683-ad517792-4c38-403d-96ef-78463e3d0817.png)

if you need to change a link, simply using ![image](https://user-images.githubusercontent.com/10260832/221163358-a7bb8553-7cfa-49fd-901a-15de750ae2ab.png). I also use it to deleted links.

If this was help... please close the issue :)

PR #2983, opened 2026-09-08
Fix popup placement and immediate dismissal on multi-monitor setups

hyperdreamer:
### Summary of Changes

This PR fixes two interrelated defects causing popup menus (menu bar dropdowns, context menus, combobox popups) to render at incorrect coordinates and close immediately upon mouse release on multi-monitor setups:

1. **Retain FlatPopupFactory for Look and Feel lifecycle**:
   - In commit `65445ebe98`, Freeplane unconditionally reverted `FlatPopupFactory` to Swing's standard `PopupFactory` in `FrameController.java`.
   - Standard Swing `PopupFactory` recycles heavyweight `JWindow` instances across different displays without verifying that the window's `GraphicsConfiguration` or DPI scaling matches the target monitor.
   - FlatLaf 3.6 natively manages the lifecycle of `FlatPopupFactory` (installing it on `FlatLaf.initialize()` and restoring the previous factory on `FlatLaf.uninitialize()`). Retaining FlatLaf's factory ensures popups are screen-aware and properly scaled across mixed-resolution displays.

2. **Resynchronize X11 frame coordinates across window manager transitions**:
   - Under X11 with window managers such as KWin (observed with OpenJDK 11–21), AWT's top-level window coordinate cache (`getLocationOnScreen()`) can fall out of sync with actual window geometry after window maximize, unmaximize, or unmaximized cross-monitor window moves.
   - When coordinates are stale, menus and dialogs calculate offsets against the wrong screen origin (opening thousands of pixels away), and mouse release events register outside the popup's perceived bounding box, triggering immediate dismissal.
   - `ApplicationViewController.java` installs an X11-gated `FrameResynchronizer` that:
     - Derives native window placement and decoration extents via bundled JNA (`_NET_FRAME_EXTENTS` via LP64 long format).
     - Filters out transient intermediate window manager geometry during animations by verifying that dimensions match within tolerance before updating coordinates.
     - Synchronizes AWT's cached frame bounds on startup, state changes (maximize/restore), and window moves using bounded deferred retries.
     - Strictly gates all native logic behind `isX11Toolkit()`, ensuring Windows, macOS, and native Wayland environments are unaffected.

### Testing Done
- Tested on a 3-monitor setup (`DP-1` 1920x1080, `DP-3` 3200x1800 primary, `HDMI-1` 1920x1080) under X11/KWin with Java 11 and Java 21.
- Verified menu dropdowns and context menus open directly under their invokers and remain open upon mouse release across all monitors.
- Verified maximize, unmaximize, and normal window moves across monitor boundaries correctly update popup anchor coordinates.
- Passed unit tests in `ApplicationViewControllerTest.java` and full `:freeplane:test` suite.

### DIFF
--- a/freeplane/src/main/java/org/freeplane/core/ui/components/FrameResynchronizer.java
+++ b/freeplane/src/main/java/org/freeplane/core/ui/components/FrameResynchronizer.java
@@ -1,0 +1,466 @@
+package org.freeplane.core.ui.components;
+
+import java.awt.Component;
+import java.awt.Frame;
+import java.awt.GraphicsConfiguration;
+import java.awt.GraphicsDevice;
+import java.awt.GraphicsEnvironment;
+import java.awt.Rectangle;
+import java.awt.Toolkit;
+import java.awt.event.ComponentEvent;
+import java.awt.event.ComponentListener;
+import java.awt.event.WindowAdapter;
+import java.awt.event.WindowEvent;
+
+import javax.swing.JFrame;
+
+/**
+ * Repairs the stale AWT window coordinate cache on the X11 toolkit so that
+ * heavyweight FlatLaf popup windows (menus, combo box dropdowns) open at the
+ * correct screen position after the window manager changes the frame geometry
+ * (maximize, unmaximize, or arbitrary move). OpenJDK 11 with KWin/X11 keeps
+ * reporting the pre-change origin via {@link java.awt.Component#getLocationOnScreen()}
+ * long after the native window moved, which misplaces popups and makes them
+ * dismiss on release.
+ *
+ * <p>Install once per top-level {@link JFrame} that hosts popup menus; the
+ * resynchronizer acts only when the toolkit is X11 and the cached coordinates
+ * actually differ from the native geometry.</p>
+ */
+public final class FrameResynchronizer extends WindowAdapter implements ComponentListener {
+	private static final String X11_TOOLKIT_CLASS_NAME = "sun.awt.X11.XToolkit";
+	private static final int RESYNCHRONIZATION_TOLERANCE = 4;
+
+	/**
+	 * Installs a {@link FrameResynchronizer} on the given frame. The listener
+	 * must be added before the frame is shown so that startup-maximized windows
+	 * are covered by {@code windowOpened} as well.
+	 */
+	public static void install(final JFrame frame) {
+		final FrameResynchronizer frameResynchronizer = new FrameResynchronizer(frame);
+		frame.addWindowListener(frameResynchronizer);
+		frame.addWindowStateListener(frameResynchronizer);
+		frame.addComponentListener(frameResynchronizer);
+	}
+
+	static boolean needsFrameResynchronization(final boolean isX11Toolkit, final int windowState,
+			final Rectangle frameBounds, final Rectangle targetBounds) {
+		return isX11Toolkit && (windowState & Frame.ICONIFIED) == 0
+				&& frameBounds != null && targetBounds != null
+				&& (frameBounds.x != targetBounds.x || frameBounds.y != targetBounds.y);
+	}
+
+	static boolean needsNormalWindowResynchronization(final boolean isX11Toolkit, final int windowState,
+			final Rectangle frameBounds, final Rectangle nativeBounds) {
+		if (!isX11Toolkit
+				|| (windowState & (Frame.MAXIMIZED_BOTH | Frame.ICONIFIED)) != 0
+				|| frameBounds == null || nativeBounds == null) {
+			return false;
+		}
+		// Only trust the native geometry while the window sizes agree. A size
+		// difference indicates the window manager is still placing or resizing the
+		// window, and acting on such transient geometry makes the window shrink.
+		if (Math.abs(frameBounds.width - nativeBounds.width) > RESYNCHRONIZATION_TOLERANCE
+				|| Math.abs(frameBounds.height - nativeBounds.height) > RESYNCHRONIZATION_TOLERANCE) {
+			return false;
+		}
+		return Math.abs(frameBounds.x - nativeBounds.x) > RESYNCHRONIZATION_TOLERANCE
+				|| Math.abs(frameBounds.y - nativeBounds.y) > RESYNCHRONIZATION_TOLERANCE;
+	}
+
+	static boolean needsMaximizedFrameResynchronization(final boolean isX11Toolkit, final int windowState,
+			final Rectangle frameBounds, final Rectangle screenBounds) {
+		return (windowState & Frame.MAXIMIZED_BOTH) == Frame.MAXIMIZED_BOTH
+				&& needsFrameResynchronization(isX11Toolkit, windowState, frameBounds, screenBounds);
+	}
+
+	static boolean needsNormalFrameResynchronization(final boolean isX11Toolkit, final int oldState,
+			final int newState, final Rectangle frameBounds, final Rectangle normalBounds) {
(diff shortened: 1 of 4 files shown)


hyperdreamer:
Hi @dpolivaev,

To provide full context on **Commit 1** (`Retain FlatPopupFactory for multi-monitor popup placement`) and how it relates to upstream commit `65445ebe98` (*"Avoid using FlatLaf popup factory because it does not work stable with JetBrains JDK"*), here is the technical background and analysis:

---

### 1. What was the original conflict between FlatLaf’s PopupFactory and JetBrains JDK?
In earlier versions of JetBrains Runtime (JBR), custom native window transparency and Projector/Wayland rendering patches in JBR's `java.desktop` conflicted with FlatLaf's translucent per-pixel alpha drop-shadow windows (`DropShadowPopup`), which `FlatPopupFactory` creates for heavyweight popups. On certain older JBR versions on Linux and Windows, this caused visual artifacts, popup flickering, or focus issues when opening menus. Commit `65445ebe98` was added as a quick workaround to force Swing back to `basicPopupFactory = new PopupFactory()`.

---

### 2. How `65445ebe98` interacted with FlatLaf
In `FrameController.setLookAndFeel()`, the override was placed inside the `if (!lookAndFeelSet)` fallback branch:
```java
if (!lookAndFeelSet) {
    ...
    if (PopupFactory.getSharedInstance().getClass().getName().equals("com.formdev.flatlaf.ui.FlatPopupFactory"))
        PopupFactory.setSharedInstance(basicPopupFactory);
}
```
However, on normal startup when FlatLaf is initialized, `FlatLaf.initialize()` installs `FlatPopupFactory` directly:
```java
oldPopupFactory = PopupFactory.getSharedInstance();
PopupFactory.setSharedInstance(new FlatPopupFactory());
```
As a result, FlatLaf still installed `FlatPopupFactory` on standard launches, but subsequent look-and-feel transitions or classloader reloads stripped it away. When replaced with `basicPopupFactory`, standard Swing recycled heavyweight `JWindow` instances across different displays without checking whether the `GraphicsConfiguration` or DPI scale matched the target monitor, causing misplaced popups on secondary screens.

---

### 3. What does FlatLaf recommend for JetBrains Runtime?
FlatLaf (which is actively tested against JBR) notes that if a specific environment or JBR build has issues with translucent popup shadows, the recommended solution is:
```java
UIManager.put("Popup.dropShadowPainted", Boolean.FALSE);
```
Disabling drop shadows turns off translucent window layering while preserving `FlatPopupFactory`'s multi-monitor screen detection, DPI scaling, and coordinate translation.

---

### 4. Will reverting `65445ebe98` cause instability?

We evaluated this carefully:
1. **Modern FlatLaf 3.6**: Freeplane now uses FlatLaf 3.6 (in `freeplane/build.gradle`), which includes upstream fixes for JBR 17/21, Linux Wayland focus, and multi-monitor setups that did not exist when the earlier shadow issues were encountered.
2. **Standard OpenJDK / Zulu / Temurin**: On standard runtimes, `FlatPopupFactory` is completely stable and essential for correct multi-monitor placement.
3. **Root cause of popup dismissal**: Our live debugging confirmed that menus closing upon mouse release on Linux X11 was caused by AWT's cached top-level window coordinates falling out of sync (addressed by Commit 2), rather than a defect in `FlatPopupFactory`.

---

### 5. Optional defense-in-depth for JBR
In practice with FlatLaf 3.6, simply removing the override in `FrameController` works cleanly. If you would prefer extra defense-in-depth for JBR without introducing configuration overhead, we can easily add:
```java
if (System.getProperty("java.vendor", "").toLowerCase(Locale.ROOT).contains("jetbrains")) {
    UIManager.put("Popup.dropShadowPainted", Boolean.FALSE);
}
```
inside `FrameController.configureFlatLookAndFeel()`.

dpolivaev:
I want to trust your analysis and I would like to accept your fix. However I would rather prefer not adding the new static methods to application view controller, is does not feel right there. I would prefer having a dedicated class for it.

But unfortunately there is more to check:
org.freeplane.view.swing.map.NodeTooltipManager.showTipWindow() uses popup menus for showing freeplane tooltips, and if I remember correctly it was one of the reasons I switched to basic PopupFactory instead of fancy L&F versions.

Can you confirm that Freeplane Tooltips still work with your solution?
Unfortunately they have pretty complex lifecycle because I wanted to let them show JEditorPane inside of JScrollPane so that user can copy their content.

hyperdreamer:
Hi @dpolivaev — both points are addressed, and I've updated the PR.

**1. Dedicated class (your first request)**

The X11 resynchronization now lives in its own class: `org.freeplane.core.ui.components.FrameResynchronizer`, with a single public entry point `FrameResynchronizer.install(JFrame)`. `ApplicationViewController` only calls `FrameResynchronizer.install(frame)` (a 2-line change) and keeps its persisted-bounds code. The unit tests moved to `FrameResynchronizerTest`. Commit 2 now adds the new class + test; Commit 1 is unchanged (the 5-line deletion that keeps FlatLaf's `PopupFactory`).

**2. Tooltips (your second request)**

I verified both the toolbar-button reproducer from your `65445ebe98` comment and the custom node tooltip path (`NodeTooltipManager.showTipWindow()` → `JPopupMenu` → `FreeplaneTooltip` → `JEditorPane` in `JScrollPane`).

Headful results:

- **Linux/X11** (OpenJDK 11 and JBR 21.0.11): node tooltips render, stay open while hovered, wheel-scroll, drag-select + Ctrl+C reaches the clipboard, and dismiss on outside click.
- **Windows 11** (JBR 21.0.6 / 21.0.7 / 21.0.11), hovering toolbar buttons and nodes:
  - official **Freeplane 1.12.11 with its bundled JBR**: 112 tooltips captured, 0 background-only;
  - patched build + FlatLaf 3.6: 83 + 113 tooltips, 0 background-only;
  - upstream baseline (basic `PopupFactory`): 11 lightweight tooltips, 0 empty;
  - patched + FlatLaf 3.7.2: 11 lightweight tooltips, 0 empty;
  - dark theme (`FlatDarkLaf`): 34 tooltips, 0 empty;
  - node-tooltip deep checks (3.6, 3.7.2 and dark): render, scroll, select+copy, dismiss all pass;
  - the File menu popup renders.

  That is 345 heavyweight + 22 lightweight tooltips, **zero background-only popups**, including your exact 1.12.11 reproducer.

  To be precise about the limits: the Windows runs were in an RDP session (software rendering, so JBR's GPU pipeline is not exercised), on JBR 21.x only, and the toolbar combobox dropdown could not be opened by synthesized input in that session — that is the one popup type I could not cover there.

**3. On the FlatLaf side of the JBR instability**

The symptom you described matches FlatLaf issue #1019 ("macOS with JetBrains Runtime: sometimes empty popups"), fixed in FlatLaf **3.6.1** by painting popup content immediately when the popup window becomes visible. Later FlatLaf releases added related popup fixes: #1037 (a tooltip in an inactive window activating it), #1029 (no longer reuse popup windows for menus), #1071 (Windows 10 popup painting) and #1082 (macOS popups on resize). Freeplane still pins FlatLaf 3.6, so none of these are in the tree — which is likely why you saw it at the time.

This PR deliberately stays on FlatLaf 3.6, so it only restores `FlatPopupFactory` without touching the dependency. If you want, I can follow up with a separate PR bumping FlatLaf (3.6.2 minimum for the reuse/activation fixes, or 3.7.2 latest; all still Java 8 bytecode) — tell me which you prefer.

dpolivaev:
@hyperdreamer many thanks, a separate PR with flatlaf update were nice if you could also test that is works.

